PR #26, opened 2014-12-23
Simplify further benchmark names

Chris00:


c-cube:
Sorry, didn't check this. What's the point of removing names exactly? Currently you can run all benchmarks under "batch.array" using `run_benchs.native -p batch.array`.

Chris00:
You have to compare with the current output to see the point.  IIRC, this is to avoid duplicate names that do not help select benchmarks anyway.  But it looks like you have changed the presentation already.

PR #30, opened 2015-01-01
Added name and dev fields in the OPAM file for local pinning.

vbmithr:
Uselful to do "opam pin add ." in the source directory.


c-cube:
thanks!

PR #32, opened 2015-01-01
fold_while implementation for CCList and CCArray (fixes #29)

### LINKED ISSUE #29: introducing fold_until convenience function
Option 1:

```
let rec fold_until (f: 'b -> 'a -> 'b * bool) (acc:'b) = 
  function | [] -> acc | e::l -> let acc, cont = f acc e in if not cont then acc else 
   fold_until f acc l;;
val fold_until : ('b -> 'a -> 'b * bool) -> 'b -> 'a list -> 'b = <fun>                                                                             
```

Pros: can accumulate and terminate on last step.
Cons: more complicated return value of the accumulator function

Option 2:
```
 let rec fold_until (f: 'b -> 'a -> 'b option) (acc:'b) = 
   function | [] -> acc  
   | e::l -> match f acc e with None -> acc | Some acc -> fold_until f acc l;;
val fold_until : ('b -> 'a -> 'b option) -> 'b -> 'a list -> 'b = <fun>                                                                             
```

Pros: simple accumulator with use of options
Cons: can't accumulate on last step

In either case, I would like to add this function to CCList, CCArray, Gen, and where-ever else it makes sense. So, do you think this function merits an addition to the library? And which signature do you prefer?

struktured:
I was looking for existing tests on these modules but couldn't find any. The tests I wanted to add were something like this:

```OCaml
(* Should return 2 *)
CCList.fold_while (fun acc b -> if b then acc+1, `Continue else acc, `Stop) 0 [true;true;false;true]

(* Should return 2 *)
CCArray.fold_while (fun acc b -> if b then acc+1, `Continue else acc, `Stop) 0 (Array.of_list [true;true;false;true])                                                                                                                      
```

### DIFF
--- a/src/core/CCArray.ml
+++ b/src/core/CCArray.ml
@@ -53,6 +53,11 @@ module type S = sig
   val foldi : ('b -> int -> 'a -> 'b) -> 'b -> 'a t -> 'b
   (** fold left on array, with index *)
 
+  val fold_while : ('a -> 'b -> 'a * [`Stop | `Continue]) -> 'a -> 'b t -> 'a
+  (** fold left on array until a stop condition via [('a, `Stop)] is
+      indicated by the accumulator
+      @since NEXT_RELEASE *)
+
   val iter : ('a -> unit) -> 'a t -> unit
 
   val iteri : (int -> 'a -> unit) -> 'a t -> unit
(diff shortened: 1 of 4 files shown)


c-cube:
First, you should rebase your branch.

You're right, tests are welcome. I use [qtest](http://batteries.vhugot.com/qtest/), so you can simply write:

```ocaml
(*$T
  foo = bar
*)
```

and `make test` should run it (check the count before and after). I'm surprised you found no such test, there are already several ones in `CCList.ml` and `CCArray.ml`.

struktured:
Rebased and added tests- I didn't expect the tests to be inlined with the implementation, hence I didn't find them.

c-cube:
Very nice. I think inline tests make it much easier to write tests, don't you?

Thanks for the good work.

struktured:
Yes, I do like the convenience of having the tests near the code but most frameworks encourage separate compilation units so it caught me off guard. Also I wish it wasn't expressed as a comment but instead via ppx or something. 

Otherwise I am now a qtest fan!

c-cube:
Agreed, ppx are cleaner, but that would break compatibility with < 4.02.

PR #35, opened 2015-01-25
Add experimental rose tree.

Emm:


src/misc/roseTree.ml
```diff
@@ -0,0 +1,168 @@
...
+    string_of_value root_value |> Format.pp_print_string formatter;
+    Format.pp_force_newline formatter ();
+    print_children root_children ""
+  in
+  print_root tree;
+  Format.pp_print_flush formatter ()
+
+module Zipper = struct
+
+  type 'a parent = { left_siblings: ('a tree) list; value: 'a; right_siblings: ('a tree) list }
```
c-cube: I prefer the following style:

```ocaml
type 'a parent = {
  left_siblings : 'a tree list;
  value : 'a;
  right_siblings : 'a tree list;
}
```

src/misc/roseTree.ml
```diff
@@ -0,0 +1,168 @@
...
+  let tree zipper = zipper.tree
+
+  let left_sibling zipper =
+    match zipper.lefts with
+    | []    -> None
+    | lefts ->
+      let rev_lefts = List.rev lefts in
+      Some {
+        tree = List.hd rev_lefts ;
+        lefts = List.tl rev_lefts |> List.rev ;
```
c-cube: pattern matching! :)
`List.tl` and `List.hd` should almost never be used...

src/misc/roseTree.mli
```diff
@@ -0,0 +1,115 @@
+(**
+   {1 Rose Tree}
+
+   A persistent, non-lazy tree where each node may have an arbitrary number of
+   children.
```
c-cube: good documentation, just please add:

`@since NEXT_RELEASE`

tests/test_RoseTree.ml
```diff
@@ -0,0 +1,600 @@
+open Containers_misc
+open OUnit
+
+let format_node value = Printf.sprintf "%d" value
+
+let string_of_tree tree =
+  let buffer = Buffer.create 1024 in
+  let formatter = Format.formatter_of_buffer buffer in
+  RoseTree.print formatter format_node tree;
+  Buffer.contents buffer
```
c-cube: you can use `CCFormat.sprintf` for this, it's more convenient.

c-cube:
Thanks a lot! I've added a few remarks, nothing serious. The test suite is really awesome. I think it might be ok to put the module directly into `containers.data` with a "CC" prefix (it looks good to me). I'll do it after merging probably.

tests/test_RoseTree.ml
Emm: How so? `CCFormat.sprintf` looks like it takes a formatting string and some arguments, but `RoseTree.print` (in its current state) needs a `formatter` (I suppose I could have it accept a `CCFormat.printer` though).

src/misc/roseTree.mli
Emm: Done

tests/test_RoseTree.ml
c-cube: Basically, with the magic of "%a"':

```ocaml
let string_of_string tree =
  CCFormat.sprintf "%a" format_node tree
```

c-cube:
I've just thought of an alternative that would allow for empty trees:

```ocaml
type 'a t = [ `Empty | `Node of 'a * 'a tree * 'a tree ]
(* node: value, first child, sibling *)
```

this encodes the list of children into the children themselves. What do you think?

Emm:
I'm not sure I understand (do you mean `Node of 'a * 'a t * 'a t`? `'a tree` is just an alias of `'a t` for the zipper). We'd also need to have right and left siblings, presumably.

Issues I see with this:
- A root with siblings suddenly becomes legit
- You'd need to use subtyping to avoid inserting `Empty in the tree
- I don't see a use case for representing empty trees that `'a t option` doesn't fulfill (apart from coherence with CCKTree)

Thoughts?

c-cube:
Right, there might be no root, indeed. Well your solution is probably the most simple, let's keep it :)

I'll need to add an export to `CCKTree` also, indeed (e.g. for export to graphviz).

PR #38, opened 2015-02-04
Elements in an LRU cache should be unique

samoht:
If it is not the case, the following code fails miserably:

```
let t = make 2
let () =
  set t 1 1;
  set t 1 1;
  set t 2 2;
  assert (get t 1 = 1)
```

because in that case the element returned by Hashtbl.find is *not* the
first bucket, which is the oldest one.

### DIFF
--- a/src/data/CCCache.ml
+++ b/src/data/CCCache.ml
@@ -269,7 +269,7 @@ module LRU(X:HASH) = struct
   let set c x y =
     let len = H.length c.table in
     assert (len <= c.size);
-    if len = c.size
+    if len = c.size || H.mem c.table x
       then replace_ c x y
       else insert_ c x y
 


samoht:
Note: there is still a bug lurking around with more complex key duplications, I'll look at it tomorrow. Would be good to write the invariants you had in mind when writing this datastructure somewhere (and ideally, would be good to have unit-tests to check that these invariants hold :p)

c-cube:
I just added a few tests, but since `LRU` isn't exposed it's not that easy to test invariants using `get` and `set`. With the way it's used currently, I don't think `set t 1 1; set t 1 1` would ever occur since `set` is only called on cache misses.

In general, the way it should work in my mind (I will write more details in the `.ml` file if you think it helps):

- when looking for a key that isn't in the table, the key/value is added in most recent position (possibly removing the key in least recent position if the cache is full)
- if the key is present, return its value, remove the key from queue and put it back in most recent position. This is dealt with using a doubly-linked circular list.

samoht:
Thanks! Adding that explanation to the `ml` could be indeed useful. And specifying that you should call `set` with an already existing key should be useful as well.

PR #41, opened 2015-02-20
Update README.md test instructions

struktured:
The existing instructions did not work. I found that running `make tests` will tell you to do `ocaml setup.ml -configure --enable-tests` if you didn't before, but the instructions I changed it to works on master and seem more to the point than the warning given after running `make tests`.

In general I found it somewhat confusing on how I was supposed to enable the tests as there seems to be several ways to go about it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ branch `stable` it is not necessary.
 
     $ make
 
-To build and run tests (requires `oUnit`, `qtest`, and `check`):
+To build and run tests (requires `oUnit`, `qtest`, and `qcheck`):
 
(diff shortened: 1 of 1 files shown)


c-cube:
Maybe it's not clear enough, I admit my focus was not on developer-oriented documentation. You need to look at _oasis to find which switches (and corresponding libraries, e.g. qtest and ounit) are required for --enable-tests to be meaningful. I'd gladly merge a PR that improves this part of the readme.
-- 
Simon

c-cube:
I was on a phone, didn't see it was a PR. Thanks!

PR #42, opened 2015-02-23
CCRingBuffer - a double sided polymorphic ring buffer

struktured:
Pretty confident on correctness, as I wrote a randomized test for every function of the signature. More interested in style and design feedback. 

Performance wise, there is an argument to be made that `blit_from_bounded` should use blit operations to batch the writes, but I don't think it's critical for the first release of this module.


c-cube:
Merged manually. Thanks a lot, great work!

c-cube:
Also changed a bit the interface and removed the `Array.Foo` modules that were mere polymorphic arrays.

struktured:
I thought those Array.Foo versions were more efficient when explicitly implemented, per the [lexifi blog](http://www.lexifi.com/blog/about-unboxed-float-arrays), but it's very possible I misunderstood it.

c-cube:
The blog post is a proposal, currently there is no efficiency gain but for Bytes.

PR #44, opened 2015-05-30
Have better default opening flags for CCIO.with_{in, out}

jpdeplaix:


src/io/CCIO.ml
```diff
@@ -134,7 +134,7 @@ let read_all ?(size=1024) ic =
   with Exit ->
     Bytes.sub_string !buf 0 !len
 
-let with_out ?(mode=0o644) ?(flags=[Open_creat]) filename f =
```
c-cube: why trunc by default? Is it because there also is `with_out_a`?

jpdeplaix:
Because open_out does it and append is not the same as trunc

c-cube:
thanks!

PR #50, opened 2015-10-13
New CCPrint.char function

Gbury:
It seems strange to not have a function to print characters individually, so there it is.


c-cube:
Nice, you should add `(** @since NEXT_RELEASE *)` in the .mli though.

Gbury:
Done.

c-cube:
thanks!

PR #54, opened 2015-11-11
Uniform implementation for CCRandom.split_list

Octachron:
This pull request modifies the implementation of `CCRandom.split_list` to make the underlying distribution uniform. All tuples (x_1,..,x_n) satisfying the constraint ∑ x_k = i should now appear with the same probability.

To test this property, I have also added a statistical uniformity test. Like any statistical test, this test has a non-zero false positive rate; but these false positives should be quite rare with the confidence interval that I have chosen (every few years assuming one test by day). 

I have also added a `sample_without_rejection` function since it can be quite useful and was needed for the implementation of `split_list`. The current implementation use rejection sampling, which works well when the number of sample is small compared to the cardinal of the sampled set.

### DIFF
--- a/src/core/CCRandom.ml
+++ b/src/core/CCRandom.ml
@@ -192,3 +192,31 @@ let (<*>) f g st = f st (g st)
 let __default_state = Random.State.make_self_init ()
 
 let run ?(st=__default_state) g = g st
+
+let uniformity_test ?(size_hint=10) k rng st =
+  let histogram = Hashtbl.create size_hint in
+  let add x = let n = try Hashtbl.find histogram x with Not_found -> 0 in
+    Hashtbl.replace histogram x (n + 1) in
+  let () =
+    for _i = 0 to ( k - 1 ) do
+      add @@ rng st
+    done in
+  let cardinal = float_of_int @@ Hashtbl.length histogram in
+  let kf = float_of_int k in
+  (* average number of points assuming an uniform distribution *)
+  let average = kf /. cardinal in
+  (* The number of points is a sum of random variables with binomial distribution *)
+  let p = 1. /. cardinal in
+  (* The variance of a binomial distribution with average p is *)
+  let variance = p *. (1. -. p ) in
+  (* Central limit theorem: a confidence interval of 4σ provides a false positive rate
+     of 0.00634% *)
+  let confidence = 4. in
+  let std = confidence *. ( sqrt @@ kf *. variance ) in
+  let predicate _key n acc =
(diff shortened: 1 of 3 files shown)


src/core/CCRandom.mli
```diff
@@ -145,3 +153,11 @@ val (<*>) : ('a -> 'b) t -> 'a t -> 'b t
 
 val run : ?st:state -> 'a t -> 'a
 (** Using a random state (possibly the one in argument) run a generator *)
+
+(** {6 Random generator testing } *)
```
c-cube: I would prefer, here, a `(**/**)` section, because this test is not really a part of the library. Additionally, a proper unit test (using qtest) would be nice, but I can add it myself later.

src/core/CCRandom.ml
```diff
@@ -78,35 +78,50 @@ let replicate n g st =
...
-      assert (len = len1+len2);
-      if i = len
-      then aux len1 ~len:len1 (aux len2 ~len:len2 acc)
-      else
-        let i1, i2 = _split (i-len) st in
-        aux (i1+len1) ~len:len1 (aux (i2+len2) ~len:len2 acc)
-  in
-  try Some (aux i ~len []) with SplitFail -> None
+  if i >= len then
+    let xs = sample_without_replacement (len-1) (int_range 1 @@ i-1) st in
```
c-cube: isn't the list already sorted, since you use a `Set`? It looks a bit wasteful to sort it again if it's the case.

c-cube:
Thanks for the patch, it looks good overall — I only have a few minor questions, but this looks like it will be merged.
The uniformity test might be very useful for several other generators, so I'd like more tests in this module.

src/core/CCRandom.ml
Octachron: You are right, this sort is completely wasteful here. Fixed.

src/core/CCRandom.mli
Octachron: Ok for a `(**/**)` section here. Concerning unit tests, did you mean something more in deep than [this short one](https://github.com/c-cube/ocaml-containers/pull/54/files#diff-f4ae21cd7c4ce81d67547aeac4ce2209R220)?

c-cube:
I forgot to ask: did you add yourself in `AUTHORS.adoc`?

Octachron:
I did not. Should I ?

c-cube:
Yes, and then I will merge.

Octachron:
Done. If needed, I could also squash the minor fixes.

c-cube:
That's fine.  Thanks again!

PR #55, opened 2015-11-22
join / both applicative functions for CCError

struktured:
```join``` and ```both``` are convenience functions which exist in Jane Street's version of ```Result``` and ```Or_error```- I thought they would also be nice to have in ``CCError`.


c-cube:
Good idea, thanks. Please just add `@since NEXT_RELEASE` in the mli doc :)

struktured:
ok, done.

PR #60, opened 2016-03-09
Make CCPersistentHashtbl.S.merge more general.

johanneskloos:
This patch brings the merge function of CCPersistentHashtbl
in line with the merge functions of other maps
(Map, BatMap from batteries, CCWBTree). In particular, its signature
changes from the restrictive
```merge: (key -> 'a option -> 'a option -> 'a option) -> 'a t -> 'a t -> 'a t```
to a more general
```merge: (key -> 'a option -> 'b option -> 'c option) -> 'a t -> 'b t -> 'c t```

### DIFF
--- a/AUTHORS.adoc
+++ b/AUTHORS.adoc
@@ -13,3 +13,4 @@
 - Guillaume Bury (guigui)
 - JP Rodi
 - octachron (Florian Angeletti)
+- Johannes Kloos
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by c-cube** [comment]
Indeed, that's a good idea. Thanks for the patch, would you mind adding your name to `AUTHORS.adoc` before I merge?

**Comment by c-cube** [comment]
Thanks!

**Comment by rgrinberg** [comment]
Just a heads up, I think that core's interface is slightly stronger for this particular function. Rather than having `'a option -> 'b option` core does

```ocaml
[` Left of 'a
| `Right of 'b
| `Both of 'a * 'b ]
```

This is slightly superior because the two optional values being `None` is impossible (if I've interpreted what this function does correctly)

c-cube:
Oh yes, I remember having seen that on #OCaml recently. That would be actually nice… But then we would need to have the same in `CCMap` (which would be a good thing).

PR #69, opened 2016-05-17
Improve CCVector.equal

copy:


c-cube:
Indeed, the old implementation is a bit stupid. Thanks!

src/core/CCVector.ml
```diff
@@ -243,12 +243,22 @@ let append_list a b = match b with
 *)
 
 let equal eq v1 v2 =
-  let n = min v1.size v2.size in
+  v1.size == v2.size
```
bluddy: Why is == used here?

c-cube: I noticed it but it's harmless, since the size is an integer.

bluddy: I guess it doesn't matter in this case, but it does bother me since it's implementation dependent. Could it have issues under js_of_ocaml?

c-cube: oh, good question. Anyway I'll fix it on master, since this is merged.

Issue #75, opened 2016-08-21
CCString.Split.left doesn't split correctly when separator is longer than 1


copy:
```ocaml
utop # CCString.Split.left ~by:"bb" "abbc";;
- : (string * string) option = Some ("a", "bc")
```

I believe this should yield `Some ("a", "c")`

copy:
On a slightly related note, it seems that the inline tests of `CCString.cppo.ml` and other cppo files aren't being run.

c-cube:
indeed, the tests in `.cppo` files are not run. Thanks for the report.

PR #76, opened 2016-08-21
Fix CCString.Split.{left,right} (#75)

copy:


c-cube:
Thanks for the fix, merged.

c-cube:
Fixed version being released!

PR #79, opened 2016-10-04
Fixes map_or

little-arhat:
~default arg for `map_or`, as well as function, should not depend on the type of error case.


c-cube:
Thank you!

Issue #84, opened 2016-10-25
Prepare for 1.0: big cleanup/simplification


c-cube:
Preparing version 1.0, with some breaking changes. The goal of this release is not to add significant new features, but to improve consistency and focus on the core features and modules (in particular, droping or reducing the scope of some of the experimental modules).

The working branch is [prepare-1.0](/c-cube/ocaml-containers/tree/prepare-1.0) .

**NOTE**: please ask if you want something changed/added to the list!

## Todo list

- [x] more consistent labels (in particular, replace `~or_` by `~default`)
- [x] label functions (`iter'`, `map'`, …) for `CCList` and the others?
- [x] doc (#85)
- [x] remove `int_`, `bool_`, etc. in `CCOrd`
- [x] make `CCIO.File.walk` work properly with symlinks, even broken
  * [x] deal with errors and files that do not actually exist
  * ~~[ ] add `CCIO.File.is_symlink`~~
  * ~~[ ] add `CCIO.File.deref : ?recursive:bool -> string -> string`~~
  * ~~[ ] make `CCIO.File.walk_item = [`Dir | `File | `Symlink] (breaking compat)~~
- [x] much simpler S-expression library using `Lexing` (see `nunchaku`)
- [x] rewrite `bitfield` (much simpler version, almost functor-free, `Bit_set`)
- [x] improve `CCFormat` (see #82)
  * [x] `CCFormat.of_to_string : ('a -> string) -> 'a t`
  * [x] `CCFormat.const : 'a t -> 'a -> unit t` (to hide the type `'a`)
  * [x] `CCFormat.text` to mimick `Format.pp_print_text`
  * [x] `CCFormat.return : (unit,…) format4 -> unit` (for strings without arguments)
  * [x] `CCFormat.some : 'a t -> 'a option t` (prints nothing if `None`, same as arg if `Some x`)
  * [x] remove `start/stop` arguments?
  * [x] make `sep` a argument-free formatter, default `",@ "`, using `Format.fprint out "%(%)" sep`
        to print it in combinators
        **OR** use `unit t` with `return: arg-free formatter -> unit t` and `sp = return ",@ "` or sth
- [x] add `CCOrd.Infix`
- [x] remove `CCPrint` (just use format)
  * [x] use `pp` as pretty-printer everywhere, and remove `Buffer.t` printers
- [x] remove `containers.advanced` (refer to `olinq`)
- [x] remove bigarray library:
[...]

PR #89, opened 2016-12-16
threads/CCLock: add try_with_lock to wrap Mutex.try_lock

dsheets:
I've added `CCLock.try_with_lock` as I've found it useful when attempting to access one of many locked resources without blocking. There is no other way that I've found to query the lockedness of a mutex and, if unlocked, acquire it atomically.

I've also updated the documentation for `get` as I initially thought it disregarded the mutex and enabled one to access the data directly. I had attempted to use this in conjunction with `CCLock.mutex` and `Mutex.try_lock` to work-around the lack of `CCLock.try_with_lock` however this caused a deadlock as the mutex had already been acquired when I called `get`.

I'm happy to work with you to get this functionality merged so let me know if there are changes you'd like to see.

Thanks!

### DIFF
--- a/src/threads/CCLock.ml
+++ b/src/threads/CCLock.ml
@@ -35,6 +35,18 @@ let with_lock l f =
   assert_equal 10 (get l)
 *)
 
+let try_with_lock l f =
+  if Mutex.try_lock l.mutex
+  then
+    try
+      let x = f l.content in
+      Mutex.unlock l.mutex;
+      Some x
+    with e ->
+      Mutex.unlock l.mutex;
+      raise e
+  else None
+
 module LockRef = struct
   type 'a t = 'a lock
   let get t = t.content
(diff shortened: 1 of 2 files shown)


c-cube:
Hi! Thanks for the PR. Note that I'm (slowly) working on version 1.0 of containers, see #84 for the roadmap, and I will be removing the `thread` library in it. Since you use it, I will move it to another repository, probably under another name (`extthread`?).

That being said, your PR is a nice addition, I will merge it as soon as you add `@since NEXT_RELEASE` to the .mli :-)

dsheets:
Interesting. Ok, `extthread` or `threadx` or something sounds good.

I couldn't find any other simple lock/blocking queue/thread pool implementations in the community but I'm sure they exist deep inside some other library or application. Anyway, for my application, preliminary measurements show about 2x end-to-end latency reduction when using your thread code vs lwt.

Thanks!

c-cube:
Yeah, I myself have some thread-specific code in some other projects -_-.

PR #94, opened 2017-01-24
Fix with_acquire: release a non locked mutex is UB

mseri:
This is segfaulting (on ubuntu 16.04 and 16.10) when trying to release the unlocked mutex in the release stage (in general this is an undefined behavior AFAIK).

### DIFF
--- a/src/threads/CCSemaphore.ml
+++ b/src/threads/CCSemaphore.ml
@@ -53,14 +53,13 @@ let release m t =
 *)
 
 let with_acquire ~n t ~f =
-  Mutex.lock t.mutex;
-  acquire_once_locked_ n t;
+  acquire n t;
   try
     let x = f() in
-    release_once_locked_ n t;
+    release n t;
     x
   with e ->
-    release_once_locked_ n t;
+    release n t;
     raise e
 
 (*$R


c-cube:
Oh right, the mutex is unlocked after `acquire`. Thanks for the fix. Can you write a minimal test case that crashes/fails currently, so we can add it to the test suite?

Note: with containers-1.0 approaching, the threading lib will be split and will go into its own opam library (with your fix, obviously).

mseri:
Actually, for me the current inline test fails reliably both on Ubuntu and on Macs. While it is successful on CentOS. I think it may depend on how pthread is compiled but I have no proof of this.

c-cube:
I can try it on arch. Anyway it's good to have!
I'll merge soon, but remember this is going to a separate library (probably called `ext-thread` or something like that)

mseri:
Absolutely, and I agree that splitting out the threading stuff makes perfect sense. 

I really like your library, thanks!

c-cube:
An alternative possibility is  that it still lives in the same repo, but with a separate package. That would be nice, I think.

c-cube:
@mseri I'd still be interested in a small test case that shows the bad behavior :-)

mseri:
I am going to make one as soon as I can and send a new PR with it

mseri:
I said something wrong, to me the tests fail reliably because I have added some indeterminism to help triggering the release of the unlocked mutex.

If you change the current test to be
```
(*$R
   let s = create 5 in
   let n = CCLock.create 0 in
   let a = Array.init 100 (fun i ->
     Thread.create (fun _ ->
       with_acquire ~n:(1 + (i mod 5)) s
         ~f:(fun () -> Thread.delay (Random.float 0.5); CCLock.incr n)
     ) ())
   in
   Array.iter Thread.join a;
   assert_equal ~printer:CCInt.to_string 5 (get s);
   assert_equal ~printer:CCInt.to_string 100 (CCLock.get n)
 *)
```
it will pass with the merged fix, but will segfault with the previous version (on many systems).

The downside is that the test becomes quite slow when it passes.

Do you want me to make a PR and modify it?

c-cube:
No need. I tried it again with a variation of your test where I replace `Thread.delay (…)` with `Thread.yield()`. On my machine this segfaults, whereas your version does not. Can you try with `Thread.yield()`?

mseri:
Yes, much faster and fails with the old code and passes with the new one. I guess we have faound a test candidate :)

c-cube:
Ok then! 1.0 getting closer :-)

PR #105, opened 2017-03-25
Plugs some holes in CCOpt

glennsl:
This adds the following functions to CCOpt:
* `map_lazy` -- for consistency with `get_lazy`
* `or_` -- as a named alternative to `<+>` and to allow `or_lazy`
* `or_lazy` -- for consistency with `get_lazy`
* `to_result` -- for consistency with other core type conversions
* `to_result_lazy` -- for consistency with `get_lazy`
* `of_result` -- for consistency with other core type conversions

### DIFF
--- a/src/core/CCOpt.ml
+++ b/src/core/CCOpt.ml
@@ -13,6 +13,10 @@ let map_or ~default f = function
   | None -> default
   | Some x -> f x
 
+let map_lazy default_fn f = function
+  | None -> default_fn ()
+  | Some x -> f x
+
 let is_some = function
   | None -> false
   | Some _ -> true
@@ -54,9 +58,15 @@ let (<*>) f x = match f, x with
 
 let (<$>) = map
 
-let (<+>) a b = match a with
-  | None -> b
+let or_ ~else_ a = match a with
+  | None -> else_
   | Some _ -> a
+
+let or_lazy ~else_ a = match a with
+  | None -> else_ ()
+  | Some _ -> a
+
+let (<+>) a b = or_ ~else_:b a
 
 let choice l = List.fold_left (<+>) None l
 
@@ -136,6 +146,18 @@ let to_list o = match o with
 let of_list = function
   | x::_ -> Some x
   | [] -> None
+
+let to_result err = function
+  | None -> Error err
+  | Some x -> Ok x
+
+let to_result_lazy err_fn = function
+  | None -> Error (err_fn ())
+  | Some x -> Ok x
+
+let of_result = function
+  | Error _ -> None
+  | Ok x -> Some x
 
 module Infix = struct
   let (>|=) = (>|=)
--- a/src/core/CCOpt.mli
+++ b/src/core/CCOpt.mli
@@ -9,8 +9,12 @@ val map : ('a -> 'b) -> 'a t -> 'b t
 (** Transform the element inside, if any *)
 
 val map_or : default:'b -> ('a -> 'b) -> 'a t -> 'b
-(** [map_or ~default f o] is [f x] if [o = Some x], [default otherwise]
+(** [map_or ~default f o] is [f x] if [o = Some x], [default] otherwise
     @since 0.16 *)
+
+val map_lazy : (unit -> 'b) -> ('a -> 'b) -> 'a t -> 'b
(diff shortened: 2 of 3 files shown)


c-cube:
Hi, thanks for the contribution! This is going to be a useful addition to containers (you should add your name to `AUTHORS` if it's not in there already). I'm going to ask for a few surface changes and then I will merge it :-)

src/core/CCOpt.ml
```diff
@@ -54,10 +58,16 @@ let (<*>) f x = match f, x with
 
 let (<$>) = map
 
-let (<+>) a b = match a with
+let or_ b a = match a with
```
c-cube: I think it's a good idea to have this non-infix alternative, indeed. What worries me a bit is the weird order of the parameters (which I understand stems from the choice in `get_lazy`…)

src/core/CCOpt.mli
```diff
@@ -9,9 +9,13 @@ val map : ('a -> 'b) -> 'a t -> 'b t
 (** Transform the element inside, if any *)
 
 val map_or : default:'b -> ('a -> 'b) -> 'a t -> 'b
-(** [map_or ~default f o] is [f x] if [o = Some x], [default otherwise]
+(** [map_or ~default f o] is [f x] if [o = Some x], [default] otherwise
```
c-cube: good catch.

src/core/CCOpt.mli
```diff
@@ -9,9 +9,13 @@ val map : ('a -> 'b) -> 'a t -> 'b t
...
     @since 0.16 *)
 
+val map_lazy : (unit -> 'b) -> ('a -> 'b) -> 'a t -> 'b
+(** [map_lazy default_fn f o] if [f o] if [o = Some x], [default_fn ()] otherwise 
+    @since 1.2 *)
```
c-cube: I tend to use `@since NEXT_RELEASE` in these comments. It's okay, because the next release will most likely be 1.2, but still.

src/core/CCOpt.mli
```diff
@@ -118,6 +130,15 @@ val to_list : 'a t -> 'a list
 val of_list : 'a list -> 'a t
 (** Head of list, or [None] *)
 
+val to_result : 'e -> 'a t -> ('a, 'e) result
+(** @ince 1.2 *)
```
c-cube: same remark about `@since NEXT_RELEASE`. Also, you wrote `@ince`.

src/core/CCOpt.ml
glennsl: It also seems natural to use it as `a |> or_ b` (though preferably in a longer chain)

c-cube: Possible, but this must be emphasized in the documentation, or perhaps even with a labelled argument (e.g. `a |> or_ ~else_:b |> or_ ~else_:c`). What do you think?

glennsl: good idea, I'll add it

glennsl:
Fixes have been commited

src/core/CCOpt.mli
```diff
@@ -94,6 +98,14 @@ val (<$>) : ('a -> 'b) -> 'a t -> 'b t
 
 (** {2 Alternatives} *)
 
+val or_ : else_:('a t) -> 'a t -> 'a t
+(** [or_ b a] is [a] if [a] is [Some _], [b] otherwise
+    @since NEXT_RELEASE *)
+
+val or_lazy : (unit -> 'a t) -> 'a t -> 'a t
```
c-cube: sorry to nitpick again (almost there), should we add `~else_` to this function, too? For the sake of consistency with `or_`?

glennsl: no worries, I can do that

glennsl: Although that wouldn't be consistent with `get_or` vs `get_lazy`

c-cube: Yes indeed, but we must live with that. It is especially critical for `or_` because the two arguments are impossible to distinguish only with their type. My feeling is that `or_lazy` should be as close to `or_` as possible…

glennsl:
Added `~else_` to `or_lazy` too. Also changed the doc comments to reflect the labels

c-cube:
Thanks! :-)

PR #114, opened 2017-04-18
Add a tail-recursive implementation of List.combine

### LINKED ISSUE #108: List.combine is not tail-rec
We probably want that in containers.

LemonBoy:
Closes #108.

### DIFF
--- a/AUTHORS.adoc
+++ b/AUTHORS.adoc
@@ -19,3 +19,4 @@
 - Malcolm Matalka (`orbitz`)
 - David Sheets (@dsheets)
 - Glenn Slotte (glennsl)
+- @LemonBoy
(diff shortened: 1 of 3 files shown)


c-cube:
Thanks for the contribution! I have a few comments to do.

- please add yourself to `AUTHORS.adoc` :-)
- I'd rather have `@since NEXT_RELEASE` than 1.2 (just in case)
- can you look at `map` and the likes to see how to combine (no pun intended) a direct-style recursion and a tail-recursion?
- It would be nice to add an entry to the benchmarks (`benchs/run_benchs.ml`, module `L`) to compare the various implementations. I can do it myself though.

I'll be happy to merge this once you make the changes :-)

PR #116, opened 2017-04-18
wip: change API of CCBV (see #113)

c-cube:
keep track of changes for bitvectors.

PR #117, opened 2017-04-19
Size tracking bitvector

rleonid:
A couple of notes:
- I didn't fully understand why the original representation used only 62 bits. I changed it to use 63. Maybe there was some confusion about the hamming distance/count bits code? But I believe that it also works for negative values.
- I changed slightly the behavior or `of_list`, `length` and `resize`.
- In general, the more that I worked on this the more that I'm puzzled about the actual usefulness of bitvectors that resize. I understand that it would be convenient, but for most use cases, the user is tracking a known set of states. In these cases I would imagine it an error for things to somewhat silently change size. It also leads to weird cases where `let i = 10000000 in get t i`  will succeed but then `set t i` might allocate a big chunk of memory. There could would seem more straight forward with fixed sizes and I'm tempted to use such a subset in my internal project.

Here are some micro benchmarks:

Creation

| Name     | Runs @ Samples | Time R^2 | Time/Run |            95ci | Cycls/Run | mWd/Run |
|---------:|---------------:|---------:|---------:|----------------:|----------:|--------:|
| Old:10   |    69757 @ 815 |     1.00 |   1.43us | -0.01us +0.01us |    4.28kc | 146.00w |
| Old:100  |     6517 @ 576 |     1.00 |  15.00us | -0.06us +0.07us |   45.01kc | 233.00w |
| Old:1000 |      582 @ 325 |     1.00 | 155.70us | -0.85us +1.13us |  467.09kc | 246.00w |
| New:10   |    74045 @ 821 |     1.00 |   1.35us | -0.01us +0.01us |    4.04kc |  74.00w |
| New:100  |     6582 @ 577 |     1.00 |  14.77us | -0.07us +0.08us |   44.32kc |  74.00w |
| New:1000 |      587 @ 326 |     1.00 | 155.06us | -0.73us +0.79us |  465.15kc |  74.00w |

Union

| Name     | Runs @ Samples | Time R^2 |     Time/Run |                95ci | Cycls/Run |    mWd/Run |
|---------:|---------------:|---------:|-------------:|--------------------:|----------:|-----------:|
| Old:10   |    83428 @ 833 |     0.98 |   1_202.80ns |   -30.65ns +51.24ns |    3.61kc |    619.00w |
| Old:100  |     7410 @ 589 |     0.98 |  13_180.18ns | -209.43ns +341.70ns |   39.54kc |  6_739.00w |
| Old:1000 |      686 @ 343 |     1.00 | 133_154.77ns | -405.34ns +415.45ns |  399.44kc | 67_939.00w |
| New:10   |   121743 @ 871 |     0.99 |     822.18ns |   -14.10ns +23.16ns |    2.47kc |    619.00w |
| New:100  |    10897 @ 628 |     0.99 |   9_041.62ns |  -88.02ns +106.60ns |   27.12kc |  6_739.00w |
| New:1000 |      987 @ 382 |     0.99 |  92_881.11ns | -639.35ns +855.09ns |  278.63kc | 67_939.00w |

Difference

| Name     | Runs @ Samples | Time R^2 |    Time/Run |                95ci |   Cycls/Run |     mWd/Run |
|---------:|---------------:|---------:|------------:|--------------------:|------------:|------------:|
| Old:10   |      324 @ 258 |     1.00 |    285.11us |     -1.11us +1.22us |    855.30kc |   1_276.01w |
| Old:100  |        81 @ 81 |     1.00 |  3_146.65us |   -24.91us +31.65us |  9_439.57kc |  13_966.00w |
| Old:1000 |        26 @ 26 |     1.00 | 31_455.14us | -151.50us +148.36us | 94_362.01kc | 140_866.00w |
| New:10   |    86814 @ 837 |     0.97 |      1.12us |     -0.01us +0.02us |      3.37kc |     619.00w |
| New:100  |     8020 @ 597 |     0.99 |     12.25us |     -0.21us +0.30us |     36.74kc |   6_739.00w |
| New:1000 |      746 @ 352 |     1.00 |    122.61us |     -0.60us +0.62us |    367.80kc |  67_939.00w |

Happy to discuss and document and test further.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -23,7 +23,7 @@ Containers is:
 - A usable, reasonably well-designed library that extends OCaml's standard
   library (in 'src/core/', packaged under `containers` in ocamlfind. Modules
   are totally independent and are prefixed with `CC` (for "containers-core"
-  or "companion-cube" because I'm megalomaniac). This part should be
+  or "companion-cube" because I'm a megalomaniac). This part should be
   usable and should work. For instance, `CCList` contains functions and
   lists including safe versions of `map` and `append`. It also
   provides a drop-in replacement to the standard library, in the module
--- a/src/data/CCBV.ml
+++ b/src/data/CCBV.ml
@@ -451,6 +451,14 @@ let diff in_ not_in =
   diff_into ~into not_in;
   into
 
+(*$T
+  diff (of_list [1;2;3])    (of_list [1;2;3])   |> to_list = [];
+  diff (of_list [1;2;3])    (of_list [1;2;3;4]) |> to_list = [];
+  diff (of_list [1;2;3;4])  (of_list [1;2;3])   |> to_list = [4];
+  diff (of_list [1;2;3])      (of_list [1;2;3;400]) |> to_list = [];
+  diff (of_list [1;2;3;400])  (of_list [1;2;3])     |> to_list = [400];
+*)
+
 let select bv arr =
   let l = ref [] in
   begin try
--- a/src/data/CCBV.ml
+++ b/src/data/CCBV.ml
@@ -1,79 +1,116 @@
 
-(* This file is free software, part of containers. See file "license" for more details. *)
-
 (** {2 Imperative Bitvectors} *)
 
-let __width = Sys.word_size - 2
+let __width = Sys.word_size - 1
 
-(* int with [n] ones *)
-let rec __shift bv n =
-  if n = 0
-  then bv
-  else __shift ((bv lsl 1) lor 1) (n-1)
+(** We use OCamls ints to store the bits. We index them from the
+    least significant bit. We create masks to zero out the most significant
+    bits that aren't used to store values. *)
+let __lsb_masks =
+  let a = Array.make (__width + 1) 0 in
+  for i = 1 to __width do
+    a.(i) <- a.(i-1) lor (1 lsl (i - 1))
+  done;
+  a
 
-(* only ones *)
-let __all_ones = __shift 0 __width
-
-type t = {
-  mutable a : int array;
-}
-
-let empty () = { a = [| |] }
-
-let create ~size default =
-  if size = 0 then { a = [| |] }
-  else begin
-    let n = if size mod __width = 0 then size / __width else (size / __width) + 1 in
-    let arr = if default
-      then Array.make n __all_ones
-      else Array.make n 0
-    in
-    (* adjust last bits *)
-    if default && (size mod __width) <> 0
-    then arr.(n-1) <- __shift 0 (size - (n-1) * __width);
-    { a = arr }
-  end
-
-(*$T
-  create ~size:17 true |> cardinal = 17
-  create ~size:32 true |> cardinal= 32
-  create ~size:132 true |> cardinal = 132
-  create ~size:200 false |> cardinal = 0
-  create ~size:29 true |> to_sorted_list = CCList.range 0 28
-*)
-
-let copy bv = { a=Array.copy bv.a; }
-
-(*$Q
-  (Q.list Q.small_int) (fun l -> \
-    let bv = of_list l in to_list bv = to_list (copy bv))
-*)
-
-let length bv = Array.length bv.a
-
-let resize bv len =
-  if len > Array.length bv.a
-  then begin
-    let a' = Array.make len 0 in
-    Array.blit bv.a 0 a' 0 (Array.length bv.a);
-    bv.a <- a'
-  end
+let __all_ones = __lsb_masks.(__width)
 
 (* count the 1 bits in [n]. See https://en.wikipedia.org/wiki/Hamming_weight *)
 let __count_bits n =
   let rec recurse count n =
     if n = 0 then count else recurse (count+1) (n land (n-1))
   in
-  if n < 0
-  then recurse 1 (n lsr 1)   (* only on unsigned *)
-  else recurse 0 n
+  recurse 0 n
+
+(*  Can I access the "private" members in testing? $Q
+  (Q.int_bound (Sys.word_size - 1)) (fun i -> __count_bits __lsb_masks.(i) = i)
+*)
+
+type t = {
+  mutable a : int array;
+  mutable size : int;
+}
+
+let length t = t.size
+
+let empty () = { a = [| |] ; size = 0 }
+
+let __to_array_legnth size =
+  if size mod __width = 0 then size / __width else (size / __width) + 1
+
+let create ~size default =
+  if size = 0 then { a = [| |] ; size }
+  else begin
+    let n = __to_array_legnth size in
+    let arr = if default
+      then Array.make n __all_ones
+      else Array.make n 0
+    in
+    (* adjust last bits *)
+    let r = size mod __width in
+    if default && r <> 0
+    then Array.unsafe_set arr (n-1) __lsb_masks.(r);
+    { a = arr; size }
+  end
+
+(*$Q
+  (Q.pair Q.small_int Q.bool) (fun (size, b) -> create ~size b |> length = size)
+*)
+
+(*$T
+  create ~size:17 true |> cardinal = 17
+  create ~size:32 true |> cardinal = 32
+  create ~size:132 true |> cardinal = 132
+  create ~size:200 false |> cardinal = 0
+  create ~size:29 true |> to_sorted_list = CCList.range 0 28
+*)
+
+let copy bv = { a = Array.copy bv.a ; size = bv.size }
+
+(*$Q
+  (Q.list Q.small_int) (fun l -> \
+    let bv = of_list l in to_list bv = to_list (copy bv))
+*)
+
+let capacity bv = __width * Array.length bv.a
 
 let cardinal bv =
   let n = ref 0 in
-  for i = 0 to length bv - 1 do
+  for i = 0 to Array.length bv.a - 1 do
     n := !n + __count_bits bv.a.(i)
   done;
   !n
+
+(*$Q
+  Q.small_int (fun size -> create ~size true |> cardinal = size)
+*)
+
+let __really_resize bv ~desired ~current size =
+  let a' = Array.make desired 0 in
+  Array.blit bv.a 0 a' 0 current;
+  bv.a <- a';
+  bv.size <- size
+
+let __grow bv size =
+  if size <= capacity bv                                   (* within capacity *)
+  then bv.size <- size
+  else                                                     (* beyond capacity *)
+    let desired = __to_array_legnth size in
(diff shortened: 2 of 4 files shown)


README.adoc
```diff
@@ -23,7 +23,7 @@ Containers is:
 - A usable, reasonably well-designed library that extends OCaml's standard
   library (in 'src/core/', packaged under `containers` in ocamlfind. Modules
   are totally independent and are prefixed with `CC` (for "containers-core"
-  or "companion-cube" because I'm megalomaniac). This part should be
+  or "companion-cube" because I'm a megalomaniac). This part should be
```
c-cube: I can fix that elsewhere ;-)

src/data/CCBV.ml
```diff
@@ -1,80 +1,117 @@
...
+  mutable size : int;
 }
 
-let empty () = { a = [| |] }
+let length t = t.size
+
+let empty () = { a = [| |] ; size = 0 }
+
+let __to_array_legnth size =
+  if size mod __width = 0 then size / __width else (size / __width) + 1
```
c-cube: I have exactly the same function in my own PR! :-)

src/data/CCBV.ml
```diff
@@ -1,80 +1,117 @@
...
-
 (** {2 Imperative Bitvectors} *)
 
-let __width = Sys.word_size - 2
+let __width = Sys.word_size - 1
+
+(** We use OCamls ints to store the bits. We index them from the
+    least significant bit. We create masks to zero out the most significant
+    bits that aren't used to store values. *)
+let __lsb_masks =
```
c-cube: interesting idea, it probably does impact perf!

c-cube:
I was not totally sure of the semantics of `lsl` w.r.t the sign bit, so I played it safe and removed it altogether, but you're right, I don't think I use anything signed anyway. So 63 bits should be safe.

src/data/CCBV.ml
```diff
@@ -1,80 +1,117 @@
...
+let __really_resize bv ~desired ~current size =
+  let a' = Array.make desired 0 in
+  Array.blit bv.a 0 a' 0 current;
+  bv.a <- a';
+  bv.size <- size
+
+let __grow bv size =
+  if size <= capacity bv                                   (* within capacity *)
+  then bv.size <- size
+  else                                                     (* beyond capacity *)
```
c-cube: stylistic nitpick: I tend to wrap such a `else` into `( … )`, to prevent any issue related to dangling else…
sth like
```ocaml
else (
   let foo = bar in
   let foo2 = bar2 in
   quux
)
```

src/data/CCBV.ml
```diff
@@ -1,80 +1,117 @@
...
+    let desired = __to_array_legnth size in
+    let current = Array.length bv.a in
+    __really_resize bv ~desired ~current size
+
+let __shrink bv size =
+  let desired = __to_array_legnth size in
+  let current = Array.length bv.a in
+  __really_resize bv ~desired ~current size
+
+let resize bv size =
```
c-cube: so you *do* keep resize, that's good.
I understand the point of making it more explicit, and I think it's a good idea.
I'm also thinking of adding `push` and `pop` to really mirror a (bit)vector…

src/data/CCBV.ml
```diff
@@ -118,34 +155,34 @@ let get bv i =
 *)
 
 let set bv i =
-  let n = i / __width in
-  if n >= Array.length bv.a
-  then resize bv (n+1);
-  let i = i - n * __width in
-  bv.a.(n) <- bv.a.(n) lor (1 lsl i)
+  if i < 0 then invalid_arg "set: negative index" else
```
c-cube: see the "dangling else" here? for an exception-throwing `if` I'd do:

```ocaml
if i < 0 then invalid_arg "foo";
…
```

but apart from that, this else should really be wrapped into `( … )`, because of the imperative `;` below.

src/data/CCBV.ml
```diff
@@ -118,34 +155,34 @@ let get bv i =
...
+    let n = i / __width in
+    let j = i mod __width in
+    if i >= bv.size then __grow bv i;
```
c-cube: like, this one. auto-indent would probably reveal the potential issue here.

src/data/CCBV.ml
```diff
@@ -276,18 +314,61 @@ let filter bv p =
     to_sorted_list bv = [2;4;6]
 *)
 
+let negate_self b =
```
c-cube: that's very good to have! neat.

src/data/CCBV.mli
```diff
@@ -21,14 +21,19 @@ val copy : t -> t
...
 val cardinal : t -> int
-(** Number of bits set *)
+(** Number of set bits. *)
 
 val length : t -> int
-(** Length of underlying array *)
+(** Length of underlying bitvector. *)
+
+val capacity : t -> int
+(** The number of bits this bitvector can store without resizing. *)
```
c-cube: need a `@since NEXT_RELEASE`

src/data/CCBV.mli
```diff
@@ -72,18 +80,34 @@ val filter : t -> (int -> bool) -> unit
 (** [filter bv p] only keeps the true bits of [bv] whose [index]
     satisfies [p index] *)
 
+val negate_self : t -> unit
```
c-cube: need `@since NEXT_RELEASE`

src/data/CCBV.mli
```diff
@@ -72,18 +80,34 @@ val filter : t -> (int -> bool) -> unit
...
+(** [negate_self t] flips all of the bits in [t]. *)
+
+val negate : t -> t
+(** [negate t] returns a copy of [t] with all of the bits flipped. *)
+
 val union_into : into:t -> t -> unit
-(** [union ~into bv] sets [into] to the union of itself and [bv]. *)
+(** [union ~into bv] sets [into] to the union of itself and [bv].
+
+    Note that [into] will grow to accammodate the union. *)
```
c-cube: "accomodate"

src/data/CCBV.mli
```diff
@@ -72,18 +80,34 @@ val filter : t -> (int -> bool) -> unit
...
 
 val inter_into : into:t -> t -> unit
-(** [inter ~into bv] sets [into] to the intersection of itself and [bv] *)
+(** [inter ~into bv] sets [into] to the intersection of itself and [bv]
+
+    Note that [into] will shrink to accammodate the union. *)
```
c-cube: same

src/data/CCBV.mli
```diff
@@ -72,18 +80,34 @@ val filter : t -> (int -> bool) -> unit
...
 val union : t -> t -> t
 (** [union bv1 bv2] returns the union of the two sets *)
 
 val inter : t -> t -> t
 (** [inter bv1 bv2] returns the intersection of the two sets *)
 
+val diff_into : into:t -> t -> unit
+(** [diff ~into t] Modify [into] with only the bits set but not in [t]. *)
+
+val diff : in_:t -> t -> t
```
c-cube: I don't really like this label. Why just no label? Difference is not ambiguous (see zarith, Int64, etc.)

c-cube:
Also: there is a benchmark suite in containers already, it would be better to use it if you can :-)
It's possible to copy the current implem into it as a "reference", I'm not afraid of code duplication in benchmarks.

c-cube:
Also, you need to add yourself to `AUTHORS`!

src/data/CCBV.ml
rleonid: Actually, this wasn't performance driven but ease of understanding. I also wanted a clearer way to have all `n` less significant bits set without baroque bit twiddling or a O(n) calculation.

I also played around with having pre-computed masks for the `1 lsl j` in `get|set|flip`'s but that isn't reliably more efficient. Specifically, it improves performance for creating small (relative to universe size) bit sets, but not for bigger cases. I think what happens is that the array used to store these values gets evicted from cache. But I didn't investigate fully.

src/data/CCBV.ml
rleonid: Interesting, though would you want to `blit` every time 63rd time that you `push` ? What's the use case?

c-cube: not, you can resize by `size p + size p / 2` (same as in some vector implementations), so you allocate exponentially less often, yet waste no more than 50% space.

src/data/CCBV.ml
rleonid: To be clear, do you prefer:
```
let set bv i =
  if i < 0 then invalid_arg "set: negative index"
  else (
    let n = i / __width in
    let j = i mod __width in
    if i >= bv.size then __grow bv i;
    Array.unsafe_set bv.a n ((Array.unsafe_get bv.a n) lor (1 lsl j))
  )
```
or
```
let set bv i =
  if i < 0 then invalid_arg "set: negative index";
  let n = i / __width in
  let j = i mod __width in
  if i >= bv.size then __grow bv i;
  Array.unsafe_set bv.a n ((Array.unsafe_get bv.a n) lor (1 lsl j))
```

My preference is for the first one.

src/data/CCBV.mli
rleonid: I can never spell this word correctly, accommodate.

src/data/CCBV.mli
rleonid: I agree that it isn't the most meaningful nor clear label. I sometimes do make mistakes and compute A \ X instead of X \ A. It is less costly to, and the mistake much readily apparent when we, compute `3-x` when you maybe meant `x-3`, as compared to mistaking set difference. But I can't think of better labels so I'll remove them.

rleonid:
Do you mind if I just grab the changes from #116 without playing `git` merging games?

c-cube:
No problem, grab changes as you like!

src/data/CCBV.mli
c-cube: For me it's quite clear without labels, I write `diff a b` like I would `a - b`.

src/data/CCBV.ml
c-cube: I'd say 1 or 3. If you prefer 1, go for it :-)

src/data/CCBV.ml
c-cube: I see. Let's keep things simple for now, if you agree — we can add a version with memoization in the benchmarks file and play with it later.

rleonid:
Ok, back to you. I'm going to pass on adding to the benchmarks as I don't have the bandwith at the moment.

rleonid:
WAIT! Just a second.

rleonid:
Ok, added a much needed test.

c-cube:
Thank you for this very nice contribution! :-)

Issue #122, opened 2017-04-28
Floored division and positive modulus


copy:
See https://en.wikipedia.org/wiki/Modulo_operation#cite_ref-7.

For example, this came up during formatting of currencies, implemented as `Printf.sprintf "%d.%02d" (n / 100) (n mod 100)`, where OCaml's `mod` produces a negative value if `n` is negative.

c-cube:
If you're willing to try and give it a shot…

PR #123, opened 2017-04-28
Implement CCInt.floor_div and CCInt.rem (#122)

copy:


src/core/CCInt.ml
```diff
@@ -37,6 +37,78 @@ let pow a b =
...
+let floor_div a n =
+  if a < 0 && n >= 0 then
+    (a + 1) / n - 1
+  else if a > 0 && n < 0 then
+    (a - 1) / n - 1
+  else
+    a / n
+
+(*$T
+  (floor_div 3 5 = 0)
```
c-cube: I'd be more comfortable with more tests with exact rounding (e.g. `floor_div (-20) 5 = -4` or sth like that)

c-cube:
This looks pretty good!

c-cube:
Awesome, thanks!

PR #150, opened 2017-09-16
Tests for CCVector.append{,_array} and fix for CCVector.append_array

copy:
This adds test for the empty vector case, and fixes `CCVector.append_array` in said case.

In celebration of containers' 1000th test, I also added myself to the authors list.


c-cube:
Very nice. Thank you, and 💯 for the 1000th test!

PR #151, opened 2017-09-18
CCFQueue bugfix for take_back_exn

quicquid:
The test "take_back empty = None" never succeeded because take_back_exn raised InvalidArgument instead.


c-cube:
Thanks! :)

PR #157, opened 2017-11-06
[WIP]  Avoid/Shadow unsafe polymorphic functions and operators 

jpdeplaix:
Polymorphic functions and operators from the standard library are somewhat unsafe to use. They can lead to unwanted exceptions, cause bugs due to differences between the naive compare and what we want to express, failures due to changes in the definitions of the type, and so on...

This PR addresses this issue by avoiding and shadowing those functions not unlike my own [ocaml-monomorphic](https://github.com/jpdeplaix/ocaml-monomorphic/) library or JaneStreet's [base](https://github.com/janestreet/base) standard library.

For now this PR is still a WIP as the tests have not been fixed yet, as this PR might be discussed extensively. I've separated different states of this proposal through different commits. Also, ```CCGraph``` is a bit tricky to change so it is going wait further discussions. This PR also fixes ```CCTrie.Make``` that required an equality function, otherwise it was failing if ```type t``` couldn't be handled by the polymorphic equality function.

I believe this PR, if accepted, might require containers to bump to 2.0 as it changes the API quite a lot and would break many users.


bluddy:
Ugh, this is such a horrible part of the language. Brings me back to the thought that OCaml is really incomplete without Modular Implicits or *something* to handle equality.

Anyway, I don't think you can make every equality/comparison argument explicit -- it's simply too much syntactic overhead.

src/core/CCEqual.ml
```diff
@@ -5,12 +5,12 @@
...
 type 'a t = 'a -> 'a -> bool
 
-let poly = (=)
+let poly = Pervasives.(=)
 
-let int : int t = (=)
-let string : string t = (=)
-let bool : bool t = (=)
-let float : float t = (=)
+let int : int t = CCInt.equal
```
c-cube: cross module dependency :(

src/core/CCListLabels.ml
```diff
@@ -1005,6 +1005,11 @@ module Assoc = struct
   *)
 end
 
+let assoc = Assoc.get_exn
```
c-cube: breaks compat, will not merge.

src/core/CCParse.ml
```diff
@@ -87,7 +87,7 @@ let next st ~ok ~err =
   else (
     let c = st.str.[st.i] in
     st.i <- st.i + 1;
-    if c='\n'
+    if CCChar.equal c '\n'
```
c-cube: in this module, again, too many cross dependencies. What's wrong with `(=)` on strings and chars?

src/threads/CCTimer.ml
```diff
@@ -6,6 +6,8 @@
 type job =
   | Job : float * (unit -> 'a) -> job
 
+open CCFloat.Infix
```
c-cube: cross module dep, etc.

c-cube:
First, thanks. It's clearly a non trivial amount of work to propose this, and I understand your point.

- I don't object to adding more explicit comparison/equality functions, adding `(=)`, `<`, … to infix modules, etc. This would not break compatibility and makes a lot of sense even in local opens.
- the part about removing all optional arguments, on the other hand, is *very* harsh on existing codebases because it breaks compatibility a lot. I don't see the ugliness of polymorphic builtins (which is real) as sufficiently bad that it's worth breaking compatibility over it.
- `CCPervasives` and, worse, the `-open` flag seem unecessary. `-open` would also need the minimum version of OCaml to be raised to 4.02, when some people still use 4.01. 
- the bugfix in `CCTrie` is neat, I'll definitely merge it/copy it somehow.

Side note, then: what's the problem with `CCGraph`? It does rely a lot on user-provided equality functions, but these (in theory) could be made mandatory?

bluddy:
IMO, a better import from Jane Street would be to break the data structures that don't work with polymorphic comparison by planting a dummy function inside them. Map could have a wrapper record that includes a dummy function, hashtable could as well etc. It'll lead to a runtime error rather than a compile one, but that's as much as we can do. This is especially important in data structures that sometimes compare correctly structurally, like hashtables.

However, this would involve breaking compatibility with the standard library and copying over all that code, so I'm not sure @c-cube wants to do this.

c-cube:
@bluddy I get the motivation, but indeed, there are two issues:

- compatibility (not the same type anymore)
- useless overhead caused by adding a pointer to every structure.

c-cube:
Ok, after another bug caused by polymorphic `min`, I'm warming up to this idea.

I propose the following:

- in the next version:
  * add `CCMonomorphic` (or similar) to src/core, which contains at toplevel int-specialized versions of all the polymorphic stuff (`min, max, +, …`).
  * add specialized operators in all relevant modules (int, float, etc.)

- in the next major version (containers 2.0):
  * `include Monomorphic` in the `Containers` module
  * make the optional arguments that would rely on poly operators mandatory (but still named). For example, `eq` arguments in `CCList.Assoc`, or `cmp` args in `CCList.sort_uniq`.

This way, code that already did not rely on poly operators, and which gave these optional arguments explicitly would not break even with containers 2.0.

bluddy:
Fair enough. It's a major fault in the language, so I guess this is inevitable.

c-cube:
Fixed by #169. if you disagree, please re-open :)

PR #158, opened 2017-11-07
Fix CCTrie.Make: Remove polymorphic comparison

jpdeplaix:
Following #157, this PR fixes a fatal use of the polymorphic equality function in ```CCTrie```.


c-cube:
Thanks! 👍

PR #161, opened 2017-12-03
Add List.remove_last

jpdeplaix:


src/core/CCList.ml
```diff
@@ -1109,6 +1109,17 @@ let remove_at_idx i l0 =
   remove_at_idx 5 [1;2;3;4] = [1;2;3;4]
 *)
 
+let rec remove_last = function
+  | [] | [_] -> []
+  | x::xs -> x::remove_last xs
```
c-cube: this is not tail-recursive, I'm a bit uncomfortable with it. Can you write something like `remove_at_idx`, or, even better, like `take` (with a fast path and a safe path)?

src/core/CCList.mli
```diff
@@ -261,6 +261,11 @@ val remove : ?eq:('a -> 'a -> bool) -> x:'a -> 'a t -> 'a t
     @param eq equality function
     @since 0.11 *)
 
+val remove_last : 'a list -> 'a list
+(** [remove_last l] removes the last element of [l]. If [l] is empty,
+    an empty list will be returned.
```
c-cube: might be good to mention it's linear time.

bluddy:
I wonder if, rather than making a new function for remove_last, it would make sense for remove_at_idx to take negative numbers -- similar to python's convention. `remove_at_idx (-1) list` would then remove the last member, but you'd also be able to easily access the second-to-last etc.

c-cube:
@bluddy that's an interesting suggestion. For coherence, indexing should also be aware of negative indices, which is not the same as what the stdlib does (but there's two sets of names anyway, `get_at_idx` and `nth`, so it would be fine).

The question of efficiency remains: `remove_at_idx (-3) list` would probably have to reverse the whole list twice, and remove the element during the second reversal. Still, it shouldn't be significantly less efficient than `remove_last`, while being more generic. I like it.

PR #169, opened 2017-12-23
Add CCMonomorphic

jpdeplaix:
Partial implementation of @c-cube's proposal: https://github.com/c-cube/ocaml-containers/pull/157#issuecomment-352435172

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -4,6 +4,7 @@ S src/iter/
 S src/sexp/
 S src/threads/
 S src/string
+S src/monomorphic
 S benchs
 S examples
 S tests
--- a/.merlin
+++ b/.merlin
@@ -20,3 +20,4 @@ PKG threads.posix
 PKG lwt
 PKG qcheck
 FLG -w +a-4-44-48-60@8
+FLG -open CCMonomorphic
--- a/_oasis
+++ b/_oasis
@@ -127,6 +127,7 @@ Executable run_benchs
   MainIs:           run_benchs.ml
   BuildDepends:     containers, qcheck,
                     containers.data, containers.iter, containers.thread,
+                    containers.monomorphic,
                     sequence, gen, benchmark
 
 Executable run_bench_hash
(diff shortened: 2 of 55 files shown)


c-cube:
This looks very good, except I don't get why monomorphic would get into its own package (and `-open` is not available in 4.01). I'd rather have `CCMonomorphic.ml` as part of src/core/, and `open CCMonomorphic` at the beginning of relevant files, for now.

c-cube:
Ok, scratch that, let's target containers 2.0!

jpdeplaix:
Ok, I have some concerns tho:
* I'm not sure of the new interface for CCGraph (with the ```~tbl``` and so on)
* I really think we should call the module ```CCPervasives``` instead of ```CCMonomorphic``` as it also handles the case where people call ```Pervasives.compare/(=)``` directly. This can happen fairly often in modules where compare is also defined for example.
* Should we also monomorphize ```Hashtbl.hash``` ?
* Should we also monomorphize ```Pervasives.(==)```/```List.assq...``` ?

c-cube:
- I'll look at the new graph interface 🙂 
- I don't think `CCPervasives` is accurate. `Pervasives` will still be available as a escape hatch for unsafe functions. `CCMonomorphic`, on the other hand, is a new, safe API.
- there is `CCHash.poly` so I suppose it'd be ok to shadow `CCHashtbl.hash`
- `==` should be hidden, not monomorphized (doesn't make sense to have both `=` and `==` on ints). A neat trick by JST is `let (==) = `Consider_using_CCEqual.physical` or whatever.
- for `List.assq` sadly I don't know :-/. I'd like not to break compat too much, but what can we do? Monomorphizing it seems a bit extreme…

jpdeplaix:
> I don't think CCPervasives is accurate. Pervasives will still be available as a escape hatch for unsafe functions. CCMonomorphic, on the other hand, is a new, safe API.

Fair enough.

> == should be hidden, not monomorphized (doesn't make sense to have both = and == on ints). A neat trick by JST is let (==) =Consider_using_CCEqual.physical` or whatever.

Done.

> edit: actually, the polymorphic hash is safe, isn't it?

Yes, but from what I heard many people usually misuse it. But I'm not really using it so I guess we can keep it as is.

> for List.assq sadly I don't know :-/. I'd like not to break compat too much, but what can we do? Monomorphizing it seems a bit extreme…

Ok fair enough. Let's keep those around for now.

PR #170, opened 2017-12-23
Allow negative indexes in CCList.remove_at_idx

jpdeplaix:
Implement suggestion by @bluddy: https://github.com/c-cube/ocaml-containers/pull/161#issuecomment-348809182

Do we want to do the same thing for every index-based functions ? (```insert_at_idx```, ```set_at_idx```, …)


c-cube:
Yes, that sounds like a good idea. Also, regarding the tests and docs: good!!

PR #176, opened 2018-01-18
Add CCList.iteri2 and CCList.foldi2

orbifx:


src/core/CCList.mli
```diff
@@ -334,9 +334,14 @@ val mapi : (int -> 'a -> 'b) -> 'a t -> 'b t
 
 val iteri : (int -> 'a -> unit) -> 'a t -> unit
 
+val iteri2 : (int -> 'a -> 'b -> unit) -> 'a t -> 'b t -> unit
```
c-cube: Looks good, but:

- add `@since NEXT_RELEASE` in comments under every new function
- add `@raise Invalid_argument when lists do not have the same length` (or something similar)

orbifx: Thanks. I amended my commit.

PR #179, opened 2018-01-26
Adding CCList.tail_opt

Fourchaux:
Small function CCList.tail_opt.
(with test & CCList.mli / CCListLabels.mli)


c-cube:
weird that this was missing…

PR #180, opened 2018-01-26
Add CCHeap.delete_all

FardaleM:


c-cube:
So the difference is that it stops as soon as it enters a subtree whose root is bigger than the element, right?

FardaleM:
Yes it is very similar to filter. Maybe I can modify the function so it remove only one occurrence of the element so it will be more useful.

What is `@since` annotation ?

c-cube:
As you said, it's impossible to implement `remove_one` properly (the element might be in several subtrees).

The annotation should be in the .mli, like:  `(** blablabla    @since NEXT_RELEASE *)` so it becomes `@since 2.1` at the next release

FardaleM:
The following function implement delete_once :
`  let  delete_once eq x h =
    let rec aux = function
      | E -> false, E
      | N(_, y, l, r) as h -> begin
        if eq x y then true, merge l r
        else begin
          if E.leq y x
          then begin
            let found_left, l1 = aux l in
            let found_right, r1 = if found_left then true, r else aux r in
            if found_right
            then true, _make_node y l1 r1
            else false, h
          end
          else false, h
        end
      end
    in
    snd (aux h)
`
Do you want me to add it ?

c-cube:
Why not indeed. Thank you!

FardaleM:
Done.

PR #184, opened 2018-02-04
CCRAL: Fixed bug in drop_tree_ / drop / take_drop

rand00:
`drop_tree` drops everything when drop-count is equal length-1:
``` ocaml
# let l = CCRAL.(drop 2 @@ add_list empty [1;2;3]) ;;
val l : int L.t = <abstr>
# let _ = CCRAL.iter (Printf.printf "%d ") l ;;
- : unit = ()
```
This PR seems to do the right thing.

### DIFF
--- a/AUTHORS.adoc
+++ b/AUTHORS.adoc
@@ -25,3 +25,4 @@
 - Fabian Hemmer (copy)
 - Maciej Woś (@lostman)
 - Orbifx (Stavros Polymenis)
+- Rand (@rand00)
--- a/src/data/CCRAL.ml
+++ b/src/data/CCRAL.ml
@@ -337,7 +337,7 @@ and drop_tree_ ~size n t tail = match t with
   | Node (_,l,r) ->
     if n=1 then append_tree_ l (append_tree_ r tail)
     else
-      let size' = size/2 in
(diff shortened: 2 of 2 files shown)


c-cube:
Thank you for the report and fix. Can you add some tests that currently fail and are fixed by this, in the qtest format? Something like:

```ocaml
(*$T
  of_list [1;2;3] |> drop 2 |> length = 1
*)
```

also, add yourself to AUTHORS if you're not there ;)

c-cube:
Thank you! 👍

rand00:
I can't say that I fully understand the effects on the underlying tree structure of the change - might be good with a review if you havn't already looked at it.

And thank you for containers (:

c-cube:
Thanks!

I'll make a second pass on it, but I think the idea is that `size` here represents the new size, which is not necessarily `2^n-1` as is normally the case because subtrees are complete binary trees.

c-cube:
I'm adding tests and digging into that anyway.

rand00:
Ah it also got more readable to me (:

c-cube:
Yes, this code was written years ago, it could use a refactoring. And it's now also more readable for me!

PR #185, opened 2018-02-04
Don't reverse twice in CCList.repeat

actionshrimp:
Looks like the implementation of CCList.repeat reversed the list at the start and at the end - this adds a test and removes one of the reversals so the result comes out the right way round.


c-cube:
Well spotted 😓 
can you add yourself to authors before I merge?

actionshrimp:
:+1: thanks!

c-cube:
Thank you! This will make it into the final 2.0 release :)

PR #202, opened 2018-02-27
Adding CCInt32 module

Fourchaux:
Seems to be a junior job.
Hoping not to make too many junior mistakes !

### DIFF
--- a/src/core/CCInt32.ml
+++ b/src/core/CCInt32.ml
@@ -1,0 +1,40 @@
+(* This file is free software, part of containers. See file "license" for more details. *)
+
+include Int32
+
+let (+) = add
+
+let (-) = sub
+
+let (~-) = neg
+
+let ( * ) = mul
+
+let (/) = div
+
+let (mod) = rem
+
+let (land) = logand
+
+let (lor) = logor
+
+let (lxor) = logxor
+
+let lnot = lognot
+
+let (lsl) = shift_left
+
+let (lsr) = shift_right_logical
+
+let (asr) = shift_right
+
+let equal (x:t) y = Pervasives.(=) x y
+
+let hash x = Pervasives.abs (to_int x)
+
+(** {2 Conversion} *)
+
+let of_string_exn = of_string
+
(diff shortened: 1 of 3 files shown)


c-cube:
Looks really nice, thank you. Can you add a `module Int32 = CCInt32` in the main containers.ml file?

(also, I'd like the opinion of @jpdeplaix on that!)

hcarty:
The the infix operators be in an `Infix` module as in  http://c-cube.github.io/ocaml-containers/last/containers/CCFloat/index.html#module-Infix?  Maybe `include Infix` in the main `CCInt32` as well.

c-cube:
@hcarty that makes sense, but it'd probably be the topic of another PR since Int64 suffers from the same flaw. I'm going to merge this as is.

src/core/CCInt32.ml
```diff
@@ -0,0 +1,40 @@
...
+
+let (lsl) = shift_left
+
+let (lsr) = shift_right_logical
+
+let (asr) = shift_right
+
+let equal (x:t) y = Pervasives.(=) x y
+
+let hash x = Pervasives.abs (to_int x)
```
jpdeplaix: Same comment as https://github.com/c-cube/ocaml-containers/pull/203/files#r171272323

jpdeplaix:
I think it should be perfect to agree to a "common" interface for those kind of module.
By common I mean:
* thinking about what we are putting into a ```Infix``` submodule
* should it be just included in the module itself ? If so, do we put everything or just a subset ?
* what sort of operation should be included in most cases ? (to_/of_ functions, that kind of thing)

I would personally be in favour of having no operators at all in the modules themselves for instance.

c-cube:
I've been updating `CCInt` to reflect the new int32/int64 interfaces. The rationale for the current state of things (having an `Infix` module + include it) is to support the two distinct use cases:

- `open Int32.Infix` in a module that will make very heavy use of their infix operators (but still use qualified names, because an unqualified `of_string` is super evil)
- very local overloading such as `Int32.(1l + 2l * 3l <> 10l)`

PR #203, opened 2018-02-28
Adding CCNativeint

Fourchaux:
Adaptation of CCInt32.

### DIFF
--- a/src/core/CCNativeint.ml
+++ b/src/core/CCNativeint.ml
@@ -1,0 +1,51 @@
+(* This file is free software, part of containers. See file "license" for more details. *)
+
+include Nativeint
+
+let equal (x:t) y = Pervasives.(=) x y
+
+module Infix = struct
+  let (+) = add
+
+  let (-) = sub
+
+  let (~-) = neg
+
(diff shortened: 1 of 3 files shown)


src/core/CCNativeint.ml
```diff
@@ -0,0 +1,51 @@
...
+
+  let (<>) = Pervasives.(<>)
+  let (<) = Pervasives.(<)
+  let (<=) = Pervasives.(<=)
+  let (>) = Pervasives.(>)
+  let (>=) = Pervasives.(>=)
+end
+include Infix
+
+let hash x = Pervasives.abs (to_int x)
```
jpdeplaix: This hash function does not give the same result as the standard ```Hashtbl.hash``` function.

@c-cube is that a problem ?

c-cube: It's not a problem, we never promised that. The hash only has to be deterministic (and hopefully not too far from uniformity)

jpdeplaix:
This comment also applies here: https://github.com/c-cube/ocaml-containers/pull/202#issuecomment-369268655

PR #206, opened 2018-03-06
Add <|> to CCLazy_list

Bronsa:
This little function is quite useful in practice

```ocaml
> to_list(default (return 1) (return 2));;
- : int list = [2]
> to_list(default (return 1) (empty));;
- : int list = [1]
```


c-cube:
Looks good! I think the `default` function should have a label on the default argument, it's prone to error otherwise.

Bronsa:
Yes makes sense -- I've added a `~default` label

PR #209, opened 2018-03-13
Comments - Style & typos fixing

Fourchaux:
Some small corrections and ... some errors/questions:

 - File src/data/CCBitField.mli    (L12)   
```#install_printer B.pp;;``` -> Unbound value B.pp 

- The lines 'status: stable' (in CCRAL.mli & CCIntMap.mli) can be deleted, no?

- La galère Monomorphic

File src/data/CCFQueue.mli (L57)
    ```[take_back_l 2 (of_list [1;2;3;4]) = of_list [1;2], [3;4]]```
 -> Error: ...an expression was expected of type int. 
What is the best way to rewrite this line to avoid this error?

File src/data/CCCache.mli   (Code sample: L21)
    ```let f' = with_cache (lru 256) f;;``` 
-> Error: This expression has type eq:'a ......

Writing ... ```(lru (=) 256) f;;```  is accepted, but is it correct?

### DIFF
--- a/src/core/CCFormat.mli
+++ b/src/core/CCFormat.mli
@@ -56,7 +56,7 @@ val flush : unit printer
     @since 1.2 *)
 
 val string_quoted : string printer
-(** Like {!CCString.print}.
+(** Similar to {!CCString.print}.
     @since 0.14 *)
 
 val list : ?sep:unit printer -> 'a printer -> 'a list printer
--- a/src/core/CCFormat.mli
+++ b/src/core/CCFormat.mli
@@ -17,7 +17,8 @@ type 'a printer = t -> 'a -> unit
 
 (** {2 Combinators} *)
 
-val silent : 'a printer (** Prints nothing *)
+val silent : 'a printer
+(** Prints nothing. *)
 
 val unit : unit printer
 (** Prints "()". *)
@@ -55,7 +56,7 @@ val flush : unit printer
     @since 1.2 *)
 
 val string_quoted : string printer
-(** Similar to {!CCString.print}.
+(** Like {!CCString.print}.
     @since 0.14 *)
 
 val list : ?sep:unit printer -> 'a printer -> 'a list printer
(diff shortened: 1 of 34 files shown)

### CODE REVIEW

File: src/core/CCFormat.mli [hunk 1]
```diff
@@ -55,7 +56,7 @@ val flush : unit printer
     @since 1.2 *)
 
 val string_quoted : string printer
-(** Similar to {!CCString.print}.
+(** Like {!CCString.print}.
```
c-cube: what is the point of this change? :/

src/core/CCFormat.mli
```diff
@@ -65,8 +66,8 @@ val seq : ?sep:unit printer -> 'a printer -> 'a sequence printer
 
 val opt : 'a printer -> 'a option printer
 (** [opt pp] prints options as follows:
-    [Some x] will become "some foo" if [pp x ---> "foo"].
-    [None] will become "none". *)
+    - [Some x] will become "some foo" if [pp x ---> "foo"].
+    - [None] will become "none". *)
```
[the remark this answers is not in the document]
Fourchaux: Instead of:
```
opt pp prints options as follows: Some x will become "some foo" if pp x ---> "foo". None will become "none".
```
ocamldoc & odoc display:
```
opt pp prints options as follows:

. Some x will become "some foo" if pp x ---> "foo".
. None will become "none".
```
Seems more readable, to me at least !

src/core/CCFormat.mli
Fourchaux: Simple homogeneity.
The goal is to replace ```'Similar to {...' or 'Analog to {...'``` with ```'Like {...'.```

c-cube: Honestly, this seems like a lot of noise for nothing. I don't care about homogeneity of wording in comments…

Fourchaux: Replacing some identical expressions with a simple one "seems like a lot of noise for nothing".

Very strange reflexion! 

But, anyway... Reverting.

c-cube:
Thank you! I appreciate your meticulosity, even when I seem to criticize :smiley:

Fourchaux:
No problemo at all.

Fun fact: 
Au jour d'aujourd'hui 'Similar to ' est unique en son genre.
Etant 'le dernier des Mohicans', 'l'exception qui confirme la règle' ou 'le village qui résiste encore et toujours à l'envahisseur' on se doit, bien sûr, de le conserver. 
(Par contre ce statut empêche l'introduction d'autres 'Similar to').

PR #211, opened 2018-03-27
Add Bijection. https://en.wikipedia.org/wiki/Bijection

orbifx:
Discussion:
http://lists.ocaml.org/pipermail/containers-users/2018-February/000042.html

### DIFF
--- a/src/data/CCBijection.ml
+++ b/src/data/CCBijection.ml
@@ -45,9 +45,10 @@ module Make(L : OrderedType)(R : OrderedType) = struct
       right = MapR.empty;
     }
 
-  let is_empty m = match MapL.is_empty m.left, MapR.is_empty m.right with
-    | l, r when l = r -> l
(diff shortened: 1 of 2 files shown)


src/data/CCBijection.ml
```diff
@@ -0,0 +1,91 @@
...
+      right : left MapR.t;
+    }
+
+  let empty = {
+      left = MapL.empty;
+      right = MapR.empty;
+    }
+
+  let is_empty m = match MapL.is_empty m.left, MapR.is_empty m.right with
+    | l, r when l = r -> l
```
c-cube: I think this should look more like:

```ocaml
let is_empty m =
  let res = MapL.is_empty m.left in
  assert (res = MapR.is_empty m.right);
  res
```

c-cube: I prefer `assert` to `Incoherence` (which is not exposed anyway)

orbifx:
Yeah, assert makes more sense. Changing it.

PR #214, opened 2018-04-19
Add CCString.{,r}drop_while

kit-ty-kate:
I'm not sure about the name though. The function taken as argument is the opposite of what `CCString.filter` is waiting for but it made sense to me.
Any comments or suggestion on another name or if they should mimic `CCString.filter` ?

### DIFF
--- a/src/core/CCString.ml
+++ b/src/core/CCString.ml
@@ -987,20 +987,23 @@ let exists p s =
   try iter (fun c -> if p c then raise MyExit) s; false
   with MyExit -> true
 
+let drop_while f s =
+  let i = ref 0 in
+  while !i < length s && f (unsafe_get s !i) do incr i done;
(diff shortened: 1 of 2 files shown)


c-cube:
That looks like a good idea. Possible names that come to me right now (not exhaustive list):

- `trim_by ~pred s`
- `trim_by ~skip_if s`
- `trim_pred f s`

kit-ty-kate:
I like `drop{l,r}` but unfortunately there is already `CCString.drop : int ‑> string ‑> string`.
I like `trim_by` but I'm not sure whether `~pred` is the right name or not. What about not having a labelled argument there ?

c-cube:
while not `{,r}drop_while f` then? would be consistent with `CCList.drop_while`

kit-ty-kate:
@c-cube sold.

c-cube:
thank you! :sparkles:

PR #218, opened 2018-04-24
Add CCOpt.guard

emillon:
Hi!

Here is an `Alternative`-style `guard` function like the one available in Haskell.
I realize that this pattern might not be very common in OCaml (though we use it), and that it might be possible to find a better name (this is very different from `CCResult.guard` for example), so let me know what you think.

Thanks!

### DIFF
--- a/src/core/CCOpt.ml
+++ b/src/core/CCOpt.ml
@@ -218,3 +218,14 @@ let flatten = function
   flatten (Some None) = None
   flatten (Some (Some 1)) = Some 1
 *)
+
+let return_if b x =
+  if b then
+    Some x
+  else
+    None
+
+(*$T
+  return_if false 1 = None
+  return_if true 1 = Some 1
+*)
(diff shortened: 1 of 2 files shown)


c-cube:
That's interesting. What do you think of `return_if : bool -> 'a -> 'a t` as a slightly more general solution? Your example wiould become `… >>= fun n -> return_if (n>0) () >>= fun () -> …`

You need to rebase, apparently :grin:

emillon:
That's pretty good because it also simplifies a bit some cases. For example we have some cases like

    let n = f x in
    guard (n > 0) >|= fun () ->
    x

This can be rewritten like this, which is pretty neat:

    let n = f x in
    return_if (n > 0) x

But in most cases, the generic "comprehension" abstraction just needs `guard` (in Haskell, `[2*x | x <- [1..5], odd x]` is compiled using `guard`).

Do you think that this kind of use case (defining generic comprehensions) would work better in `olinq`?

c-cube:
Olinq is a different beast, having this in `Option` makes a lot of sense anyway. I'll accept this PR if you change to `return_if`, since we both agree it's neat :-)

emillon:
Right, let's find this magical abstraction later :)

PR #223, opened 2018-05-29
Add a leq function for each primitive types

Julow:
This function allows these modules to be passed to CCHeap.Make

I can also add it to some modules of `containers.data` or to `CCSexp`

### DIFF
--- a/src/core/CCBool.ml
+++ b/src/core/CCBool.ml
@@ -7,6 +7,8 @@ let equal (a:bool) b = Pervasives.(=) a b
 
 let compare (a:bool) b = Pervasives.compare a b
 
+let leq (a:bool) b = Pervasives.(<=) a b
+
 let negate = not
 
 type 'a printer = Format.formatter -> 'a -> unit
--- a/src/core/CCBool.mli
+++ b/src/core/CCBool.mli
@@ -8,6 +8,10 @@ type t = bool
 val compare : t -> t -> int
 (** Total ordering on booleans, similar to {!Pervasives.compare}. *)
 
+val leq : t -> t -> bool
+(** Equivalent of [(<=)]
+    Allows to be passed to {!Heap.Make} *)
+
 val equal : t -> t -> bool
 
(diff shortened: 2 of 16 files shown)


c-cube:
Thanks for the contribution!
I think it'd be cleaner to provide a `(<=)` function in primitive modules that don't have one already, and to provide a second functor `CCHeap.Make'` (for example) that would use `<=` instead of `leq` in its parameter.

I must say I use `CCHeap` sufficiently rarely that writing:

```ocaml
module H = CCHeap.Make(struct
  type t = foo
  let leq = (<=)
end)
```

doesn't seem too bad :-)

Julow:
About the `(<=)` I'd find it even better if all the operators were included into primitive modules. I have a lot of frustration with the `Infix` modules

For the `CCHeap.Make'` functor, better make it take a "comparable" module. eg:

```
module Make' (O : sig
	type t
	val compare : t -> t -> int
end) =
struct
	include Make (struct
		type t = O.t
		let leq a b = O.compare a b <= 0
	end)
end
```

c-cube:
Usually I think the `Infix` modules are also included in the module itself. If it's not the case it'd be a welcome change (stuff like `include Infix` in the primitive modules that lack it).

About `compare`: no, that's too strong. `Heap` only requires a partial order (with transitivity), not a total order like `compare`.

Julow:
`compare` is too strong but it is just for convenience. My point is making a special case that implement `leq` from `(<=)` is too simple to be worth the pollution of the Heap module.

I'm still in favor of a `Heap.Make_from_compare`. I think it add a lot of convenience to this module.
And the advantage over adding these `leq` everywhere is that it will work on modules from other library that don't implement `leq`.

c-cube:
I don't like the idea of adding `leq` everywhere just for the sake of this heap module, when `<=` does the job perfectly well and is standard.

On the `compare` side, ok, if there's a docstring that makes it clear it's for convenience then go for it! :-)

PR #224, opened 2018-05-29
Fix Int32 and Int64 operators are not visible

Julow:
Operators are defined in the Infix module.
This module was included in the .ml file but not in the .mli file

I also made an (maybe) unrelated change to CCEqual. I can remove it


c-cube:
Good! :+1: 
Please add yourself to `AUTHORS` and I'll merge

Julow:
Thanks !

PR #225, opened 2018-05-29
Add CCHeap.Make_from_compare

Julow:
Allow to pass module implementing `compare` without implementing `leq`

There is some discussion here: https://github.com/c-cube/ocaml-containers/pull/223#issuecomment-392784680

`TOTAL_ORD` is not the usual name.

I added a test but I did not succeed in running it

### DIFF
--- a/src/core/CCHeap.ml
+++ b/src/core/CCHeap.ml
@@ -15,6 +15,15 @@ module type PARTIAL_ORD = sig
   (** [leq x y] shall return [true] iff [x] is lower or equal to [y]. *)
 end
 
+module type TOTAL_ORD = sig
+  type t
+  val compare : t -> t -> int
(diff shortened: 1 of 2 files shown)


src/core/CCHeap.ml
```diff
@@ -403,3 +412,22 @@ module Make(E : PARTIAL_ORD) : S with type elt = E.t = struct
          pp_elt out x)
       h
 end
+
+module Make_from_compare(E : TOTAL_ORD) =
+struct
```
c-cube: this would be more concise as `module Make_from_compare(E:TOTAL_ORD) =
   Make(struct type t = E.t … end)`

no need for inclusion :-)

Julow: This syntax is pretty cool

Issue #216, opened 2018-04-20
Have a containerLabels module


bluddy:
The StdLabels module opens the stdlib with labelled modules as the default. It would be nice to allow the same functionality with Containers.

madroach:
Given a labelled module interface, could one not automatically generate a non-labelled version of it by simply dropping all labels?
Not as simple, but still not too difficult would be the task of generating a non-labelled implementation file out of the labelled interface file.

Given this interface and implementation
``` ocaml
module type CCArrayLabels = sig
  val make : int -> 'a -> 'a array
  val get : 'a array -> int -> 'a
  val set : 'a array -> int -> 'a -> unit
  val fill : 'a array -> pos:int -> len:int -> 'a -> unit
end
module CCArrayLabels :CCArrayLabels = ArrayLabels
```
the task would be to generate this interface and implementation:
``` ocaml
module type CCArray = sig
  val make : int -> 'a -> 'a array
  val get : 'a array -> int -> 'a
  val set : 'a array -> int -> 'a -> unit
  val fill : 'a array -> int -> int -> 'a -> unit
end

module CCArray : CCArray = struct
  open CCArrayLabels
  let make = make
  let get = get
  let set = set
  let fill a b c d = fill a ~pos:b ~len:c d
end
```

Looks like its easily scriptable to me or am I missing something?

c-cube:
I'm not sure it's so easy to script. Also, even in non-label modules, there are some labels when we're not tied by retrocompat.

madroach:
What would be the preferred layout of interfaces?
* ``Array.Labels``
* ``ArrayLabels``
* ``Labels.Array``
* ``Containers.Labels``
* other ?

madroach:
So if I manage to hack up some script would this be considered for inclusion and the obvious modules be labelled?

c-cube:
Another problem is the docstrings, where there are code samples like `Foo.map f l` instead of `Foo.map ~f l`. I'm also feeling a bit uneasy removing things like `CCList.mli` and have them autogenerated…

madroach:
#227 implements my suggestion to autogenerate unlabelled interfaces.
It automatically fixes the docstrings.

c-cube:
so maybe now we can do something like a `containersLabels.ml` file with:

```ocaml
include Containers
module String = CCStringLabels
module Array = CCArrayLabels
(* … *)
```
What do y'all think?

c-cube:
[...]

PR #227, opened 2018-06-10
Generate unlabelled interfaces from labelled ones automatically

madroach:
* add unlabel.ml which parses interfaces and removes labels in function declarations and docstrings. This is a quick hack, but it does the job.
 * an attribute [@keep_label] can be added to the labelled arguments to mark labels like ~cmp which should stay in the "unlabelled" interface
 * While augmenting the labelled interfaces I stumbled across some inconsistencies between the currend labelled / non-labelled interfaces. I used the labelling that made most sense to me and added a "FIXME" comment.
 * Maybe we should break backwards compatibility at some point and name the comparison arguments to the sorting functions in CCArrayLabels.ml ~cmp instead of ~f ?

Before doing further cleanup and adding more labelled modules I would wish for some feedback on the general approach I'm taking here.

Please note that the unlabelled interfaces are generated with
[``mode promote``](https://jbuilder.readthedocs.io/en/latest/jbuild.html#modes). This may help to ease @c-cube's concerns about autogenerating an interface like ``CCList.mli``

It would help a lot in maintaining something like a ContainerLabels module #216 .


src/core/CCArray.mli
```diff
@@ -13,7 +21,11 @@ type 'a printer = Format.formatter -> 'a -> unit
 
 (** {2 Arrays} *)
 
-include module type of struct include Array end
+(**/**)
+external make_float : int -> float array = "caml_make_float_vect" (* compat *)
+(**/**)
+
+include module type of Array
```
c-cube: not sure what this change is about? the `struct include end` trick is pretty convenient…

src/core/CCArray.mli
```diff
@@ -57,28 +69,30 @@ val length : _ t -> int
 (** [length a] returns the length (number of elements) of the given array [a]. *)
 
 val fold : ('a -> 'b -> 'a) -> 'a -> 'b t -> 'a
-(** [fold f acc a] computes [f (... (f (f acc a.(0)) a.(1)) ...) a.(n-1)],
+(** [fold f init a] computes [f (... (f (f init a.(0)) a.(1)) ...) a.(n-1)],
```
c-cube: why change this?

src/core/CCArray.mli
```diff
@@ -130,97 +147,103 @@ val sort_ranking : ('a -> 'a -> int) -> 'a t -> int array
...
 
-val lookup_exn : cmp:'a ord -> 'a -> 'a t -> int
-(** [lookup_exn cmp x a] is like {!lookup}, but
-    @raise Not_found if the key [x] is not present. *)
+val lookup_exn : cmp:('a ord [@keep_label]) -> 'a -> 'a t -> int
+(** [lookup_exn ~cmp key a] is like {!lookup}, but
+    @raise Not_found if the key [key] is not present. *)
 
-val bsearch : cmp:('a -> 'a -> int) -> 'a -> 'a t ->
+val bsearch : cmp:(('a -> 'a -> int) [@keep_label]) -> 'a -> 'a t ->
```
c-cube: I'm afraid the `keep_label` attribute might break on 4.02 :-/
It seems the inline attribute is already problematic…

We may need some kind of AST preprocessor which depends on OCaml's version?

src/core/CCArrayLabels.mli
```diff
@@ -106,10 +108,12 @@ val reverse_in_place : 'a t -> unit
 (** [reverse_in_place a] reverses the array [a] in place. *)
 
 val sorted : f:('a -> 'a -> int) -> 'a t -> 'a array
+    (* FIXME: better label this ~cmp ?? *)
```
c-cube: maybe in 3.0, but it doesn't seem super important to me

src/core/CCList.mli
```diff
@@ -1,3 +1,12 @@
+(*
```
c-cube: is this an effect of the promotion, btw?

src/core/CCListLabels.mli
```diff
@@ -66,7 +66,10 @@ val fold_map : f:('acc -> 'a -> 'acc * 'b) -> init:'acc -> 'a list -> 'acc * 'b
     @since 0.14 *)
 
 val scan_left : f:('acc -> 'a -> 'acc) -> init:'acc -> 'a list -> 'acc list
-(** @since 2.2 *)
+(** [scan_left ~f ~init l] returns the list [[init; f init x0; f (f init x0) x1; ...]]
+    where [x0], [x1], etc. are the elements of [l].
+    @since 1.2, but only
+    tsince 2.2 with labels *)
```
c-cube: typo on `tsince`

src/core/jbuild
```diff
@@ -1,4 +1,18 @@
 
+(rule
+  ((targets (CCArray.mli))
+   (action (run ${ROOT}/src/unlabel.exe ${<} ${@}))
+   (deps (CCArrayLabels.mli ${ROOT}/src/unlabel.exe))
+   (mode promote)
```
c-cube: "promote" will copy the generated file back to the source tree, correct?

src/unlabel.ml
```diff
@@ -0,0 +1,192 @@
+let slurp_file file =
+  let ch = open_in file in
+  let buf = Buffer.create (min 1024 (in_channel_length ch)) in
+  try
+    while true do Buffer.add_channel buf ch 4096 done;
```
c-cube: this might be buggy on old versions of OCaml, I remember propositing a patch for it… The last chunk, if smaller than 4096, might be dropped. To be checked…

src/unlabel.ml
```diff
@@ -0,0 +1,192 @@
...
+  let mli_text = slurp_file labelled_mli_filename in
+  let lexbuf = Lexing.from_string mli_text in
+  Location.init lexbuf labelled_mli_filename;
+  let labelled_ast = Parse.interface lexbuf in
+  let labelled_functions = ref [] in
+  let rec iterator =
+    let open Ast_iterator in
+    let open Parsetree in
+    { Ast_iterator.default_iterator with
+        value_description = fun iterator vdesc ->
```
c-cube: We should also remove attributes for `ocaml <= 4.02`, I think. There's a `attributes` field in the iterator where we could filter the attributes specific to this PR (and possibly others like `[@inline]`?).

src/unlabel.ml
```diff
@@ -0,0 +1,192 @@
...
+
+    (* hacks *)
+    let unlabelled_mli =
+      Buffer.to_bytes unlabelled_mli |> Bytes.unsafe_to_string
+      |> Str.(global_replace (regexp_string labelled_name) unlabelled_name )
+    in
+
+    let out = open_out unlabelled_mli_filename in
+    output_string out (
+      "(*\n *\n *\n *      AUTOGENERATED FROM " ^
```
c-cube: this output is a bit verbose for my taste, a single-line comment should be enough

src/unlabel.ml
```diff
@@ -0,0 +1,192 @@
...
+              ds_end
+            end
+            else raise Not_found
+          with Not_found -> start
+        in
+        unlabel start tail
+    in unlabel 0 labelled_functions;
+
+    (* hacks *)
+    let unlabelled_mli =
```
c-cube: not sure what this is about? Is it for updating the docstrings?

src/unlabel.ml
```diff
@@ -0,0 +1,192 @@
...
+    |> List.flatten
+  in
+
+  begin (* check whether positions are in order *)
+    List.map (fun (_, {Lexing.pos_cnum}) -> pos_cnum) labels
+    |> List.fold_left (fun a b -> assert (a <= b); b) 0
+    |> ignore
+  end;
+
+  Format.pp_open_vbox fmt 0;
```
c-cube: That's pretty weird. I think it'd be better to use a single `Format.fprintf fmt "@[<v>…@]@." with sub-printers, and avoid the `\n` in format strings (use `@,` for inserting breaks instead).

src/core/CCArray.mli
madroach: Since ``CCArray.mli`` is autogenerated I did not actually make any changes to it. Any changes result from inconsistencies betweed ``CCArray.mli`` and ``CCArrayLabels.mli``.
What is this trick of ``struct include end`` about? Should this be done the same way in ``CCArrayLabels.mli``? In that case it needs to be changed there.

src/core/CCArray.mli
madroach: bacause its called ``~init`` in ``CCArrayLabels.mli``. The whole point of this proposal is to reduce the maintanance costs of labelled interfaces, so any inconsistencies should be erased. Please note that in this case backwards compatibility is not affected. Changing the label name in ``CCArrayLabels.mli`` would break backwards compatibility.

src/core/CCArray.mli
madroach: I'll test on 4.02. This label is obviously by easily removed automatically.

src/core/CCList.mli
madroach: yes. The autogenerated file is promoted from the ``_build`` dir to the ``src`` dir. You may remove it from the source dir and jbuilder will generate it automatically.

src/core/jbuild
madroach: correct. They won't be removed during clean, but automatically updated on OCaml >= 4.03.0.

src/unlabel.ml
madroach: works for me on OCaml 4.03.0. Won't be used on 4.02.*

src/unlabel.ml
madroach: unlabelled.ml does no preprocess *ASTs* like a ppx preprocessor, but *.mli*.
At least at the moment it is only supposed to generate unlabelled .mli from a labelled one. The labelled one is not changed. Would you like a preprocessor reworking all *.mli* and possibly also *.ml* ?

madroach: I'm afraid the AST won't help with removing attributes from mli, since the AST doesn't contain locations of the attributes.

src/unlabel.ml
madroach: changed this to a single line comment.

src/unlabel.ml
madroach: Removed the ``Format.`` stuff altogether. It was only needed while implementing the rewriter.

madroach:
> can we generate the files on a recent OCaml and, using promote, check them out so they also compile on 4.02? That'd be good for avoiding dealing with many versions of the parsetree in the preprocessor, and also to prevent containers from depending on compiler-libs.

That's how its done already. See the Makefile.

> we should strip the attributes we use, because 4.02 doesn't seem to like them.In a way, that also makes the preprocessor potentially useful for doing other things than unlabelling. 

If attributes need to be stripped, then all ``.mli``s would need to be preprocessed. That would mean to put the interfaces in a ``CC%%%%%.blub.mli`` and generate ``CC%%%%%Labels.mli`` and ``CC%%%%%.mli`` from it.

> updating labels in docstrings should be doable in the AST mapper, as docstrings (as of 4.03) are just attributes with ``[@@ocaml.doc "the doc here"]``.

the unlabel.ml preprocessor I wrote rewrites ``.mli``s, not ASTs. So the AST mappor won't be helpful as information is lost in the mli -> AST transformation.


> Usage of Str should be as narrow as possible.

All I'm using ``Str`` for is searching for plain strings and replacing plain strings. Is there any other library available that can provide this functionality at compile-time?

src/core/CCArray.mli
madroach: Tested and doesn't break 4.02.3.
It seems like 4.02 not only supports inline attributes for expressions, but also for type expressions :smile:

src/core/CCArray.mli
c-cube: the `struct include end` has been discussed over IRC, I think. It's useful for keeping equalities between types (i.e. `'a CCArray.t = 'a Array.t` for the typechecker).

We can change it in `CCArrayLabels.mli` indeed!

src/core/CCArray.mli
c-cube: :+1:

src/unlabel.ml
c-cube: It's not necessarily a good idea to get a general preprocessor right now, I was just thinking forward.
Do you know the difference between `Ast_iterator` and `Ast_mapper`?

src/core/CCArray.mli
madroach: done.

src/unlabel.ml
madroach: ``Ast_iterator`` is an iterator over the AST while ``Ast_mapper`` is a mapping function mapping for transforming ASTs

madroach:
Hi, before I start to work on this again and resolve the merge conflicts I'd like to know whether there still is interest in this pull request and what else will need to be improved.

UnixJunkie:
It lowers the maintenance burden so it looks quite interesting.

c-cube:
I still have an instinctive negative reaction to magic used for generating code, but this makes sense anyway (especially with `promote` so that normal compilation shouldn't need to do any generation). Is there a chance you may use `CCString` itself instead of `Str` to replace the strings? It may not be compiled yet, so there's a bootstrapping issue, but in general str is quite frowned upon these days.

I'm sorry for the lack of enthusiasm, but if you want to push this forward I'll support you and will merge if it's clean enough. It should be helpful in the future as you pointed out, so thank you.

madroach:
This works for now without ``Str``. When I find time, I may try to search for the docstrings via the AST.

madroach:
Now I believe the ``unlabel.ml`` rewriter is ready for commit.
My (small) changes to the interfaces still need some attention and design decisions made

c-cube:
I think moving to `CCArrayLabels.ml` is fine, and more consistent. Just be sure qtest picks it properly :-)

c-cube:
It looks good to me, really. I made small changes (removed obsolete dep on str, etc.) but I wonder if it doesn't always build unlabel.exe even if it doesn't need to? Which means we'd always depend on compiler-libs?

c-cube:
Merged manually after a few changes, on master.
Many thanks to @madroach for their perseverance and efforts! :sparkles:

c-cube:
This is kind of backfiring with 4.08, it seems, because `Compenv.module_of_filename` has changed its signature -_-

PR #228, opened 2018-08-31
Support using ** for exponentiation in CCInt

loxs:
This will allow writing something like `CCInt.(10 + 5 ** 2)` instead of `CCInt.(10 + (pow 5 2))`


c-cube:
Hi, thanks for contributing! That PR seems like a good idea, I'd just like you to do two things:

- add yourself to the authors file
- add `@since NEXT_RELEASE` as a comment to the new function

:-)

c-cube:
All good, thank you!

PR #229, opened 2018-10-05
Add CCResult.get_or_failwith

hongchangwu:
Useful for raising meaningful exceptions when dealing with results with string errors.

Equivalent functions include https://github.com/dbuenzli/rresult/blob/master/src/rresult.mli#L138 & https://github.com/janestreet/base/blob/master/src/result.mli#L60

### DIFF
--- a/AUTHORS.adoc
+++ b/AUTHORS.adoc
@@ -31,3 +31,4 @@
 - Christopher Zimmermann (@madroach)
 - Jules Aguillon (@julow)
 - Metin Akat (@loxs)
+- Hongchang Wu (@hongchangwu)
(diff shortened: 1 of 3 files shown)


hongchangwu:
Test failures seems to come from some randomized test:

```
# Failure: >split_list>../src/core/CCRandom.ml:229
# 
# ../src/core/CCRandom.ml:229:  run ~st:(QCheck_runner.random_state()) ( uniformity_test 50_000 (split_list 10 ~len:3) )
```

c-cube:
Excellent! Nothing to change in this PR, thank you very much.
The test fails from time to time when we're unlucky (literally)…

PR #230, opened 2018-10-05
added List.counts, related to List.count

UnixJunkie:


c-cube:
That looks like a useful function. Two remarks:

- please add yourself to the authors file
- I don't like the name, it's not explicit at all. How about something like `count_true_false`?

UnixJunkie:
all done

c-cube:
I think you included a "merge" commit instead of rebasing.

UnixJunkie:
Yes, I'm not very good at git.
Squashing is perfect.

PR #233, opened 2018-10-18
add containersLabels.ml

### LINKED ISSUE #216: Have a containerLabels module

madroach:
fixes #216

### DIFF
--- a/src/core/containersLabels.ml
+++ b/src/core/containersLabels.ml
@@ -3,11 +3,11 @@
 
 (** {1 Drop-In replacement to Stdlib} *)
 
-module Array = CCArray
-module Array_slice = CCArray_slice
+module Array = CCArrayLabels
+module Array_slice = CCArray_sliceLabels
 module Bool = CCBool
 module Char = Char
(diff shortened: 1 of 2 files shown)


src/core/containers.ml
```diff
@@ -45,9 +41,10 @@ module Ref = CCRef
 module Result = CCResult
 module Set = CCSet
 module String = CCString
-module StringLabels = CCStringLabels
 module Vector = CCVector
 module Monomorphic = CCMonomorphic
 module Utf8_string = CCUtf8_string
 
+module Labels = CCArrayLabels
```
c-cube: is this intended? :)

c-cube:
So you're generating `containers.ml` from `containersLabel.ml`? That's nice, especially for future updates. There seems to be one surprising bit but I'll merge when it's fixed.

:100:

madroach:
Well, no, its not autogenerated, but I believe this will be mainanable.

PR #236, opened 2018-11-07
Fix #235 for CCRingBuffer

### LINKED ISSUE #235: CCRingBuffer / CCVector keeps deleted values alive
I'm using a `CCRingBuffer` as a buffer for packets coming from a network to process them later. Typically there is not more than one packet in the buffer and they arrive and are consumed in regular time intervals.

The consuming functions of `CCRingBuffer` ([`take_front_exn`](https://github.com/c-cube/ocaml-containers/blob/2a97950/src/data/CCRingBuffer.ml#L367), etc.) don't overwrite the removed element in the underlying array, so the deleted value is kept alive (from the point of view of the garbage collector). Even worse, with my use case, the deleted values are often promoted to the major heap.

This also applies to `CCVector`.

I have observed this using statmemprof.

Base seems to do this correctly: https://github.com/janestreet/base/blob/209e74b/src/queue.ml#L216

copy:
A few notes:

- I added `dummy` to the signature of the Array in order to use it from `MakeFromArray`, which is a breaking change and not terribly nice. Open to better suggestions (add the `Elt` submodule to `Array`?)
- It doesn't seem to be possible to add comments to `$inject` blocks
- On my machine, `live_blocks` is 2 if deleting is properly implemented and `capacity b + 2` otherwise, so I believe the GC test is reasonably stable, but of course I'm also open to other approaches

### DIFF
--- a/src/data/CCRingBuffer.ml
+++ b/src/data/CCRingBuffer.ml
@@ -345,20 +345,6 @@ module MakeFromArray(A:Array.S) : S with module Array = A = struct
     to_buf = s && len = Bytes.length s)
   *)
 
-  let clear b =
-    b.stop <- 0;
-    b.start <- 0;
-    ()
-
-  (*$Q
-    a_str (fun s -> let s = Bytes.of_string s in \
-    let s_len = Bytes.length s in \
-    let b = Byte.create (max s_len 64) in \
-    Byte.blit_from b s 0 s_len; \
-    Byte.clear b; \
-    Byte.length b = 0)
-  *)
-
   let is_empty b = b.start = b.stop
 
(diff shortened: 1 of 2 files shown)


c-cube:
That looks good! Just a few nitpicks:

- `clear` doesn't seem to reclaim memory?
- `dummy` should have an `@since` annotation

copy:
> clear doesn't seem to reclaim memory?

It does, unless I'm misunderstanding (it calls `junk_front` `length` times).

c-cube:
My bad, I missed the fact that `skip` was doing all the work. Looks good, thanks for the contribution! :+1:

PR #242, opened 2018-12-20
[WIP] Work on documentation

FardaleM:
I may add more doc in the future. Maybe, the best plan is to merge this just before the next release.

### DIFF
--- a/src/core/CCResult.mli
+++ b/src/core/CCResult.mli
@@ -188,6 +188,9 @@ end
 (** {2 Collections} *)
 
 val map_l : ('a -> ('b, 'err) t) -> 'a list -> ('b list, 'err) t
+(** [map_l f l] apply [f] on every element of [l], and, in case of success for every
+    element, return [Ok (List.map f l)]. Otherwise, it fails, and the first error is
+    return.*)
 
 val fold_l : ('b -> 'a -> ('b, 'err) t) -> 'b -> 'a list -> ('b, 'err) t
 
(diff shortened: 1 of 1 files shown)


src/core/CCResult.mli
```diff
@@ -188,6 +188,9 @@ end
 (** {2 Collections} *)
 
 val map_l : ('a -> ('b, 'err) t) -> 'a list -> ('b list, 'err) t
+(** [map_l f l] apply [f] on every element of [l], and, in case of success for every
+    element, return [Ok (List.map f l)]. Otherwise, it fails, and the first error is
+    return.*)
```
c-cube: "returned" :slightly_smiling_face: 
also, it's not really `Ok (List.map f l)` but the list of all the `Ok`-values, right?

FardaleM: It is equivalent to `Ok (List.map f l)` but it does not call List.map. It is the list of unwrap Ok-values. But I don't know how to write it accurately and concisely.

pmetzger: Explain it verbosely (I don't currently understand what the function does) and I will edit it down.

FardaleM: I pushed a new version of this.

src/core/CCResult.mli
```diff
@@ -188,6 +188,9 @@ end
 (** {2 Collections} *)
 
 val map_l : ('a -> ('b, 'err) t) -> 'a list -> ('b list, 'err) t
+(** [map_l f \[a1; ...; an\]] applies function f to a1, ..., an , and, in case of
+    success for every elements, return the list of [Ok]-value.
```
pmetzger: Small note: `element` should be singular (to agree grammatically with `every`.)

pmetzger: Oh, and `returns` should be plural.

PR #244, opened 2019-01-21
Use GADT to discard impossible case on CCFQueue.

dinosaure:


src/data/CCFQueue.mli
```diff
@@ -10,7 +10,7 @@ type 'a printer = Format.formatter -> 'a -> unit
 
 (** {2 Basics} *)
 
-type +'a t
+type 'a t
```
c-cube: It's possible to make `'a t` covariant again, by also making `digit` covariant.

PR #249, opened 2019-02-25
ocamldebug section in the README

nilsbecker:
This may be too long already?

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,34 @@ See issues
 [#196](https://github.com/c-cube/ocaml-containers/issues/196),
 [#197](https://github.com/c-cube/ocaml-containers/issues/197)
 
+## Debugging with `ocamldebug`
+
+To print values with types defined in `containers` in the bytecode debugger,
+you first have to load the appropriate bytecode archives. After starting a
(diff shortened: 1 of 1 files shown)


c-cube:
This seems good, but tests don't pass. Does `make` work locally for you (with mdx installed)?

nilsbecker:
not really; i did `opam install mdx` then `make clean` then `make`. i get 
```
dune build   @install
dune runtest --no-buffer --force
Done: 671/680 (jobs: 2)Got an error while evaluating:
---
module M = struct let pp_int_list = Containers.(List.pp Int.pp) end
---
Characters 36-46:
Error: Unbound module Containers
         mdx alias runtest (exit 1)
(cd _build/default && /Users/nbecker/.opam/4.07.1+flambda/bin/mdx test README.md)
make: *** [test] Error 1
```

nilsbecker:
ah wait. this is more sophisticated than i imagined -- my code examples are actually being built and run? i did not think they would need to be self-contained. can i just make this code block into an inline code snippet to prevent it from being built and tested?

c-cube:
Yes, the readme is tested to prevent the code samples from becoming stale. Thank you for the patch!

PR #250, opened 2019-03-07
don't use Compenv.module_of_filename

madroach:
fix https://github.com/c-cube/ocaml-containers/pull/227#issuecomment-469780960


c-cube:
Perfect, thank you!

PR #259, opened 2019-04-29
Remove unused reference in CCIO.read_chunks

rymdhund:
It looks like the eof reference was accidentally left behind after some old refactoring


c-cube:
indeed, good catch! Thanks :tada:

PR #263, opened 2019-07-05
Fix Containers.Stdlib on OCaml 4.07

kit-ty-kate:
The Stdlib module has been introduced from OCaml 4.07 and not OCaml 4.08. As a consequence submodules from the standard library were not available anymore when you open the Containers module.


c-cube:
That was such a stupid mistake, thank you! I think I'll make a point release soon.

PR #264, opened 2019-10-07
Add functions from new Stdlib.Fun

madroach:
Add new functions from ``Stdlib.Fun``. For functions we already had, use the implementation from ``Stdlib.Fun``.
Two issues, about which I'm unsure how to address:
* This obviously only works when ``Stdlib.Fun`` is available, which it is only since OCaml 4.08. 
* Licensing: OCaml Stdlib is modified LGPL, while containers is BSD. May I quote the documentation from the Stdlib as I do in this PR, or should I just try to link to the Stdlib?

I noticed [mkshims.ml](src/core/mkshims.ml). Should I just include ``fun.ml`` from the OCaml distribution (under LGPL) and use it when OCaml <4.08, but not use it when OCaml >=4.08 ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,4 @@ env:
   - PACKAGE="containers" OCAML_VERSION="4.06" EXTRA_DEPS="base-threads base-unix"
   - PACKAGE="containers" OCAML_VERSION="4.07" EXTRA_DEPS="base-threads base-unix"
   - PACKAGE="containers" OCAML_VERSION="4.08" EXTRA_DEPS="base-threads base-unix" TESTS=false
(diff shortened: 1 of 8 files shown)


Gbury:
Couldn't it be simpler to simply include the `Fun` module (both for the implementation and the documentation) ?

madroach:
Yes, I'd like to do that. But what to do when we compile on OCaml <4.08, where there is no ``Fun`` module?

c-cube:
This seems interesting, I'll take a look tonight. I'm a bit unsure about including actual code from the stdlib (it doesn't seem hard to reimplement with the same interface) and about removing the existing functions' docs and `@since` attributes.

c-cube:
It fails on the CI when tests are enabled, any idea why?

c-cube:
superseeded by #266

PR #265, opened 2019-10-07
add CCArray.Infix

narimiran:
This is my first ever PR containing OCaml code, so please double-check everything and tell me what needs to be corrected and/or done differently.

I've used [CCList.ml](https://github.com/c-cube/ocaml-containers/blob/master/src/core/CCList.ml#L1704) and [CCList.mli](https://github.com/c-cube/ocaml-containers/blob/master/src/core/CCList.mli#L709) (links are to the relevant parts of the code) as a guidelines, but I'm not sure if this is the way to do it (e.g. everything was added manually).

Also, if there is a reason why this was not included in the first place, feel free to reject this PR.


c-cube:
Thank you, `Infix` modules are nice :+1:  :)

PR #267, opened 2019-10-21
deque: filter in place

c-cube:


src/data/CCDeque.ml
```diff
@@ -323,6 +327,107 @@ let to_rev_list q = fold (fun l x -> x::l) [] q
...
+  (* update size, compute new cell *)
+  let update_local_ n =
+    d.size <- d.size - size_cell_ n.cell;
+    let new_cell = filter_cell_ f n.cell in
+    d.size <- d.size + size_cell_ new_cell;
+    new_cell
+  in
+  let rec loop ~stop_at n : unit =
+    let n_prev = n.prev in
+    let n_next = n.next in
```
FardaleM: Why didn't you put this two definition inside the if?

src/data/CCDeque.ml
```diff
@@ -323,6 +327,107 @@ let to_rev_list q = fold (fun l x -> x::l) [] q
...
+        | Two (x,y), One z -> remove_node_ n; n_prev.cell <- Three (x,y,z)
+        | One x, Three (y,z,w)
+        | Two (x,y), Two (z,w) -> n_prev.cell <- Three (x,y,z); n.cell <- One w
+        | Two (x,y), Three (z,w1,w2) -> n_prev.cell <- Three (x,y,z); n.cell <- Two (w1,w2)
+      end;
+      loop ~stop_at n_next;
+    );
+  in
+  d.cur.cell <- update_local_ d.cur; (* special case for first cell *)
+  loop ~stop_at:d.cur d.cur.next
```
FardaleM: The side effect is that the first node will be a Three. But does not seem to be a problem.

PR #271, opened 2019-11-01
change type of CCDeque

FardaleM:
Change the definition of the type in CCDeque to remove the Zero cell.
This new type enforce one invariant.


c-cube:
thanks! :sparkles:

PR #277, opened 2019-12-04
Re-enable mdx tests

NathanReb:
We recently released `mdx.1.5.0` which contains fixes for some of the bugs you ran into so I took the liberty to open a PR re-enabling the mdx check of the README.

I also rewrote the dune rules by hand as `mdx rule` still needs to be improved a little. As a matter of fact it's going to be replaced by a proper dune integration for mdx in an hopefully near future!

Let me know what you think!

### DIFF
--- a/dune
+++ b/dune
@@ -1,7 +1,7 @@
 (rule
- (target README.md.corrected)
+ (targets README.md.corrected)
  (deps (package containers))
- (action (run ocaml-mdx test %{dep:README.md} -o %{target})))
+ (action (run ocaml-mdx test %{dep:README.md} -o %{targets})))
 
 (alias
  (name runtest)
--- a/dune
+++ b/dune
@@ -1,8 +1,8 @@
+(rule
+ (target README.md.corrected)
+ (deps (package containers))
+ (action (run ocaml-mdx test %{dep:README.md} -o %{target})))
 
-;(alias
-; (name runtest)
-; (deps README.md)
-; (action (progn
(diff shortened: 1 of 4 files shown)


dune-project
```diff
@@ -1 +1 @@
-(lang dune 1.0)
```
c-cube: why is that needed? :)

NathanReb: The new rules I wrote use some `dune.1.11` features such as the `target` field. I can try to lower the required version down a bit if you want!

c-cube: were the old rules wrong?

NathanReb: Sorry, I forgot to answer! Yeah using a `progn` in this case isn't very friendly to dune as it doesn't know about generated files. Also the `diff?` behaviour isn't the best, we've had issue with these mdx rules in the past.

dune
```diff
@@ -1,8 +1,8 @@
+(rule
+ (target README.md.corrected)
+ (deps (package containers))
+ (action (run ocaml-mdx test %{dep:README.md} -o %{target})))
```
c-cube: btw @NathanReb I get messages like that:

```
ocaml-mdx: unknown option `-o'.
Usage: ocaml-mdx test [OPTION]... FILE
Try `ocaml-mdx test --help' or `ocaml-mdx --help' for more information.
``` and I'm not sure why?

NathanReb: You need to upgrade to `mdx.1.5.0`! The `-o` option was introduced in that version.

PR #282, opened 2019-12-14
Wip ccvec

FardaleM:


src/core/CCVector.ml
```diff
@@ -378,7 +383,8 @@ let pop_exn v =
   let new_size = v.size - 1 in
   v.size <- new_size;
   let x = v.vec.(new_size) in
-  if new_size > 0 then v.vec.(new_size) <- v.vec.(0); (* free last element *)
+  (* free last element *)
+  fill_with_junk_ v.vec new_size 1;
```
c-cube: hmm it would be worth having `junk_of : 'a -> 'a` and use that instead of filling with size `1`, which is probably a lot harder to inline.

PR #284, opened 2020-01-07
Avoid catch-all handlers

copy:
Two other occurences weren't fixed because their exceptions are not documented: https://github.com/c-cube/ocaml-containers/blob/4bc01a0b8251dae4e2e5ae7648b5c03911804c23/src/unix/CCUnix.ml#L313-L314
 https://github.com/c-cube/ocaml-containers/blob/4bc01a0b8251dae4e2e5ae7648b5c03911804c23/src/core/CCIO.ml#L282

### DIFF
--- a/src/core/CCIO.ml
+++ b/src/core/CCIO.ml
@@ -222,11 +222,7 @@ let with_in_out ?(mode=0o644) ?(flags=[Open_creat]) filename f =
 let tee funs g () = match g() with
   | None -> None
   | Some x as res ->
-    List.iter
-      (fun f ->
-         try f x
-         with _ -> ()
-      ) funs;
+    List.iter (fun f -> f x) funs;
     res
 
 (* TODO: lines/unlines:  string gen -> string gen *)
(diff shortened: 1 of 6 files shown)


src/core/CCIO.ml
```diff
@@ -222,11 +222,7 @@ let with_in_out ?(mode=0o644) ?(flags=[Open_creat]) filename f =
 let tee funs g () = match g() with
```
copy: As a side-note, should this function be in CCIO?

c-cube: it's useful to apply several functions to an input stream, but agreed, it's a bit specialized.
it's there so I think we should keep it, though.

src/core/CCIO.ml
```diff
@@ -222,11 +222,7 @@ let with_in_out ?(mode=0o644) ?(flags=[Open_creat]) filename f =
...
   | None -> None
   | Some x as res ->
-    List.iter
-      (fun f ->
-         try f x
-         with _ -> ()
-      ) funs;
+    List.iter (fun f -> f x) funs;
```
c-cube: I think that makes sense, but can you add a comment to the .mli to document the behavior? (There should have been one before too, of course).

PR #286, opened 2020-01-24
examples: clean up

FardaleM:
Remove old example and fix the others


c-cube:
I feel like even the remaining example could go… into the printbox repo?

FardaleM:
lambda.ml definitely, it does not use containers at all. id_sexp.ml does not use Printbox and uses containers.sexp.

c-cube:
so we can remove lambda, I think :)

PR #219, opened 2018-04-29
Improving comments presentation

Fourchaux:
. The goal is to start EACH (humpf...) comment line showing the usage of the function. That is:
 
```
(** [foo_name arg1 arg2 ...] followed by the description .... *)
```
. Additions/modifications comments (files CCArray / CCArrayLabels / CCArray_slice).

. Small typos.

PR #291, opened 2020-03-16
Comments presentation modifications.

Fourchaux:
It is a continuation of PR https://github.com/c-cube/ocaml-containers/pull/219 .
(Almost 2 years ago. Well,well...).

Question: 
The mixture of _Pervasives.foo_ and _Stdlib.foo_ is growing.
For a next PR I propose to replace _Pervasives_ by _Stdlib_.

Do you agree?


c-cube:
I think using `Stdlib` everywhere is good, yes. Now we have shims to be able to do that even pre-4.08 :-0

PR #294, opened 2020-03-28
CCSexp: Escape empty atoms

Julow:
Hello,

In `CCSexp`, this PR escape empty atoms with quotes (`""`), otherwise `` `List [ `Atom "" ] `` would be formatted as `()`, which is `` `List [] ``.

It seems the tests were expecting that (`sexp_valid`), was this an intended behavior ?


c-cube:
well to me it made little sense to have `Atom ""` as it can never be parsed, but I imagine that `("")` is actullay what would produce that. So, good catch.

PR #301, opened 2020-04-26
CCVector.remove

FardaleM:
Rename CCVector.remove to CCVector.remove_unordered and create a new CCVector.remove that keep the order of the element in the vector.


c-cube:
I'd rather rename `remove` to avoid ambiguities (maybe `remove_and_shift`?).

FardaleM:
Done

c-cube:
perfect, thanks a lot :muscle:

PR #302, opened 2020-04-27
feat(CCVector) remove shrink_to_fit and rename shrink in keep

FardaleM:
As the memory of unused cell is now freed, I think we do not need `shrink_to_fit anymore`.
For `shrink`, I think it is usually called `keep`.


c-cube:
No, `shrink_to_fit` is still useful, for example if you want to store the vector for a while and don't want memory to be wasted.

And I don't like `keep` (to me it's more of a "filter elements given a predicate" function, like `retain`). How about `shrink_to_size : 'a t -> int -> unit` which indicates it shrinks to the given size?

FardaleM:
Ok I will try to think of better name.

PR #306, opened 2020-05-15
Add CCFloat.pi

sonologico:
A small first contribution adding something that I was missing that is present in Stdlib. Let me know if something could be improved.

### DIFF
--- a/src/core/CCFloat.ml
+++ b/src/core/CCFloat.ml
@@ -38,6 +38,8 @@ let max_finite_value = Stdlib.max_float
 
 let epsilon = Stdlib.epsilon_float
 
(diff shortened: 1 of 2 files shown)


c-cube:
it needs to be reimplemented on versions lower than 4.08 (I think? cf CI) since `Float.pi` doesn't exist on, say, 4.03.

sonologico:
In that case, do you see any downside in inlining the definition? On the Stdlib it is defined as `0x1.921fb54442d18p+1`. It's small enough shouldn't require any change.

c-cube:
I guess we can copy it, then.

PR #307, opened 2020-05-16
feat(vector): transform shrink in resize

FardaleM:
Fix name confusion between `shrink` and `shrink_to_fit` by introducing `resize` instead of `shrink`.

### DIFF
--- a/src/core/CCVector.ml
+++ b/src/core/CCVector.ml
@@ -112,7 +112,7 @@ let resize_ v newcapacity x =
   push v 0.; push v 1.;
   clear v;
   push v 0.; push v 1.; push v 7.; push v 10.; push v 12.;
-  shrink v 2;
+  truncate v 2;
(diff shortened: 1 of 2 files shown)


c-cube:
I find `resize` very confusing, as it doesn't actually resize if you pass a bigger size than the current one.

FardaleM:
> I find `resize` very confusing, as it doesn't actually resize if you pass a bigger size than the current one.

It does resize if you pass a bigger size. It is now `val resize : ('a, rw) t -> 'a -> int -> unit

c-cube:
This makes it more annoying to use when you just want to shrink.

note that rust also has [shrink](https://devdocs.io/rust/std/vec/struct.vec#method.shrink_to) but also something called `truncate` (which is closer to what we have). Maybe we could rename that to `truncate`?

FardaleM:
Should be fine now.

PR #313, opened 2020-06-19
CCFun: put infix operators in Infix module

### LINKED ISSUE #312: CCFun Infix module
Various other modules, like `CCResult` and `CCList`, enclose their infix operators within an `Infix` sub-module, making it easier to expose/open access to just the operators without cluttering the namespace with all of the module's (non-infix) functions.  It seems like it would be convenient, and good for consistency's sake, to do the same with `CCFun`?

Such a change wouldn't have to be backwards-incompatible--the `Infix` module can still be `include`d at the module top-level (just as `CCList` and `CCResult` do right now), so that the infix operators can still be accessed from `open CCFun`, and existing code relying on this behavior doesn't break.

kwshi:
Resolves #312.

### DIFF
--- a/src/core/CCFun.mli
+++ b/src/core/CCFun.mli
@@ -5,9 +5,6 @@
 
 include module type of CCShimsFun_
 
-val (|>) : 'a -> ('a -> 'b) -> 'b
-(** [x |> f] is the same as [f x]. A 'pipe' operator. *)
-
 val compose : ('a -> 'b) -> ('b -> 'c) -> 'a -> 'c
 (** [compose f g x] is [g (f x)]. Composition. *)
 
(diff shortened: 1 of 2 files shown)


src/core/CCFun.mli
```diff
@@ -79,6 +66,26 @@ val iterate : int -> ('a -> 'a) -> 'a -> 'a
...
+  (** [f @@ x] is the same as [f x], but right-associative.
+      @since 0.5 *)
+
+  val (%>) : ('a -> 'b) -> ('b -> 'c) -> 'a -> 'c
+  (** [(f %> g) x] or [(%>) f g x] is [g (f x)]. Alias to [compose]. *)
+
+  val (%) : ('b -> 'c) -> ('a -> 'b) -> 'a -> 'c
+  (** [(f % g) x] or [(%) f g x] is [f (g x)]. Mathematical composition. *)
+end
+
```
c-cube, line 88: I think you need to `include Infix` here, or `include module type of Infix`. A test is required, because we could have missed that and introduced a regression :)

kwshi: Oh wow, good catch!  I added an `include module type of Infix` in f7fbf8f0.  When you say "a test is required", do you mean I should add some here, or as something that should be done in a later PR/eventually?  :p

c-cube: it'd be nice to do it now, it can be as simple as this in the .ml:

```ocaml
(*$T
  CCFun.((succ % succ) 2 = 4)
  CCFun.((succ %> string_of_int) 2 = "3")
*)
```

kwshi: Is the `CCFun` necessary?  (I'm assuming this would go in `CCFun.ml` to begin with.)

kwshi: Nevermind--it appears that I don't actually know how qtest works.  Is there anyway to do this using a `$=` instead of `$T`?  I gave it a try, but it doesn't seem to like what I'm writing.

kwshi: Anyway, added a handful of $T tests in fee29724; happy to change them if needed!

c-cube: both would work (`$=` takes 2 arguments), this should do. Thanks for adding tests! :100:

PR #318, opened 2020-06-25
CCResult: add <$> operator

### LINKED ISSUE #317: CCResult could use a <$> operator
`CCList` has one, `CCOpt` has one, `CCResult` could use one too!

kwshi:
Resolves #317.

The one other minor change I've made is that instead of redundantly declaring infix signatures `<*>` and `>|=` once at the top-level and then once again in `Infix`, they are now imported into the top-level of the `CCResult` module signature via `include module type of Infix`.

### DIFF
--- a/src/core/CCResult.mli
+++ b/src/core/CCResult.mli
@@ -114,12 +114,6 @@ val catch : ('a, 'err) t -> ok:('a -> 'b) -> err:('err -> 'b) -> 'b
 
 val flat_map : ('a -> ('b, 'err) t) -> ('a, 'err) t -> ('b, 'err) t
 
-val (>|=) : ('a, 'err) t -> ('a -> 'b) -> ('b, 'err) t
-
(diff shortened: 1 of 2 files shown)


src/core/CCResult.mli
```diff
@@ -185,7 +174,11 @@ val both : ('a, 'err) t  -> ('b, 'err) t -> (('a * 'b), 'err) t
 (** {2 Infix} *)
 
 module Infix : sig
+  val (<$>) : ('a -> 'b) -> ('a, 'err) t -> ('b, 'err) t
```
c-cube, line 177: needs a `@since NEXT_RELEASE`  ! :)

kwshi: Just to be clear, am I literally writing the text `NEXT_RELEASE`, or substituting in the version number for the next release (`2.9`?).

c-cube: literally write `@since NEXT_RELEASE` :-)
Have a look in the makefile to see how it's removed. And next version should be 3.0 hopefully.

PR #311, opened 2020-06-14
comments modifs

Fourchaux:
Continuation of https://github.com/c-cube/ocaml-containers/pull/310

PR #319, opened 2020-06-26
Comments

Fourchaux:
Continuation of #311

### DIFF
--- a/src/core/CCOpt.mli
+++ b/src/core/CCOpt.mli
@@ -96,7 +96,7 @@ val get_lazy : (unit -> 'a) -> 'a t -> 'a
     @since 0.6.1 *)
 
 val sequence_l : 'a t list -> 'a list t
-(** [sequence_l [x1; x2; ⋯; xn]] returns [Some [y1; y2; ⋯;yn]] if
+(** [sequence_l [x1; x2; …; xn]] returns [Some [y1; y2; …; yn]] if
     every [xi] is [Some yi]. Otherwise, if the list contains at least
     one [None], the result is [None]. *)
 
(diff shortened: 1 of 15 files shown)


src/core/CCOpt.mli
```diff
@@ -23,119 +23,126 @@ val is_none : _ t -> bool
...
     @raise Invalid_argument if the option is [None]. *)
 
 val get_lazy : (unit -> 'a) -> 'a t -> 'a
-(** [get_lazy default_fn x] unwraps [x], but if [x = None] it returns [default_fn ()] instead.
+(** [get_lazy default_fn o] unwraps [o], but if [o] is [None] it returns [default_fn ()] instead.
     @since 0.6.1 *)
 
 val sequence_l : 'a t list -> 'a list t
-(** [sequence_l [x1; x2; ...; xn]] returns [Some [y1;y2;...;yn]] if
+(** [sequence_l [x1; x2; ⋯; xn]] returns [Some [y1; y2; ⋯;yn]] if
```
c-cube, line 99: do you mean to use `…` instead of `···` ? :)

Fourchaux: Euh, yes I mean to use … instead of . . . 
Is there a pb somewhere?

Note: maybe ⋯ would be better than ⋯; (no ; after ⋯)?

c-cube: you're using · instead of . in your dots (middle of the line instead of bottom).

Fourchaux: OK, I see.
Killing all these little beasts.

c-cube:
thank you! :)

PR #322, opened 2020-07-18
Stdlib.Seq

### LINKED ISSUE #254: migrate to `Iter` and add type definition `'a iter` and `{to,of}_iter` everywhere
we can also break compat and make `of/to_seq` alias to the stdlib's

- [x] use `to_std_seq` in 2.xx
- [ ] rename to `to_seq` in 3.0?

### LINKED ISSUE #257: remove `ccklist` in favor of `seq`
not necessary anymore!

- [x] deprecate it in 2.xx
- [ ] remove it in 3.0

FardaleM:
Fix #257 
Fix #254

### DIFF
--- a/src/core/CCHeap.ml
+++ b/src/core/CCHeap.ml
@@ -149,11 +149,7 @@ module type S = sig
   val size : t -> int
   (** Number of elements (linear complexity). *)
 
-  (** {2 Conversions}
-
-      The interface of [of_gen], [of_iter]
-      has changed since 0.16 (the old signatures
-      are now [add_seq], [add_gen]). *)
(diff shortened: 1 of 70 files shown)


src/core/CCHeap.ml
```diff
@@ -152,9 +151,9 @@ module type S = sig
 
   (** {2 Conversions}
 
-      The interface of [of_gen], [of_iter], [of_klist]
+      The interface of [of_gen], [of_iter]
       has changed since 0.16 (the old signatures
```
c-cube, line 155: I think this can just be dropped, no one cares about 0.16 anymore :)

PR #326, opened 2020-07-30
Comments

Fourchaux:
Continuation of https://github.com/c-cube/ocaml-containers/pull/319

Modifs comments for CCStringLabels.mli.

**Question:**
The first line of CCString.mli says:
(*  AUTOGENERATED FROM CCStringLabels.mli  *)

Still valid? No need to change CCString.mli?


FardaleM:
I don't think CCString.mli is generated from CCStringLabels.mli now. Maybe @c-cube can confirm this.

c-cube:
indeed, I just removed that comment. The generation was complicated and brittle; now there's just a test that fails if the interfaces don't match.

PR #331, opened 2020-09-21
Move definition of CCMap.update so that it is shadowed by Stdlib.Map.…

jberdine:
…update

Signed-off-by: Josh Berdine <josh@berdine.net>

### DIFF
--- a/src/core/CCMap.ml
+++ b/src/core/CCMap.ml
@@ -166,6 +166,15 @@ module Make(O : Map.OrderedType) = struct
     | None -> raise Not_found
     | Some (k,v) -> k, v
 
+  let update k f m =
+    let x =
+      try f (Some (M.find k m))
+      with Not_found -> f None
+    in
+    match x with
+      | None -> M.remove k m
+      | Some v' -> M.add k v' m
+
   include M
 
   let get = find_opt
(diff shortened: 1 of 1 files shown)


c-cube:
is it the same type? I didn't look at the stdlib's Map recently.

jberdine:
Yes, it seems that this interface is the natural one. :-)  From the stdlib map.mli:
```ocaml
    val update: key -> ('a option -> 'a option) -> 'a t -> 'a t
    (** [update x f m] returns a map containing the same bindings as
        [m], except for the binding of [x]. Depending on the value of
        [y] where [y] is [f (find_opt x m)], the binding of [x] is
        added, removed or updated. If [y] is [None], the binding is
        removed if it exists; otherwise, if [y] is [Some z] then [x]
        is associated to [z] in the resulting map.  If [x] was already
        bound in [m] to a value that is physically equal to [z], [m]
        is returned unchanged (the result of the function is then
        physically equal to [m]).
        @since 4.06.0
    *)
```
and CCMap.mli
```ocaml
  val update : key -> ('a option -> 'a option) -> 'a t -> 'a t
  (** [update k f m] calls [f (Some v)] if [find k m = v],
      otherwise it calls [f None]. In any case, if the result is [None]
      [k] is removed from [m], and if the result is [Some v'] then
      [add k v' m] is returned. *)
```

c-cube:
Good call, thank you. I also just added a test. :)

PR #334, opened 2020-11-05
Add a guard function for list comprehensions

grayswandyr:
See #333.

In principle, we could also add such functions for all monads with a zero (e.g. options). I only did it for lists however. 

Also, as there's no sensible zero for `CCResult` (the only module already featuring a `guard` function), we could keep calling the added function `guard` (as in Haskell) instead of `mguard`, if you see fit.


c-cube:
thanks!

Issue #338, opened 2020-12-08
CCSexp.parse_string crash/unexpected Invalid_argument exception


darrenldl:
As part of fuzzing [timere](https://github.com/daypack-dev/timere), cases which cause `CCSexp.parse_string` to raise `Invalid_argument("Char.chr")` exception are found.

## Manually minimised example

`\"\\N\"` where N is an integer > 255, e.g. `\"\\256\"`

A reprod is as follows:

```
let () =
  CCSexp.parse_string "\"\\256\" |> ignore
```

gives (with backtrace)

```
Fatal error: exception Invalid_argument("Char.chr")
Raised at file "stdlib.ml", line 30, characters 20-45
Called from file "src/core/CCSexp_lex.mll", line 46, characters 30-75
Called from file "src/core/CCSexp_lex.mll", line 77, characters 18-63
Called from file "src/core/CCSexp_lex.ml", line 169, characters 3-33
Called from file "src/core/CCSexp.ml", line 186, characters 30-35
Called from file "src/core/CCSexp.ml", line 229, characters 16-25
```

## Original cases from fuzzer

In case I missed something when minimising manually, following are some of the fuzzer discovered cases
- `\"\n\n<\\r9999999Y\\\"\n9\\99999999999999g9999999999999g999999g\"\n`
- `\"\n\r<\\t\023\n\203\\622222222\\\\\"\"\2032!2222\\\\\"\"`
- `\"\n\r<@t\023\n\203\\2222D2\n\r22222\01622222222222222222222222\203\\292242\222 2\\\\\">K2`
- `\"\n\r<\\t\023\n\203\\272222222\\\\\"\"\2032\0042222\\\\\"\"`
- `\"\023\n\203\\5222\n\r<\\t\023\n\203\\52222222\\\\\"2\\\216\216\216\216\216\\\\\"\216\216\216\216\216\216\216\216\216222222222222222\147`
- `\"\\722\02622222\\\\\n\r<\\\\\\\\\"\\222222222\\\\\"\"\2032!2222\\\\\"\"`

Crowbar/afl gives the same error for above inputs, e.g.

```
When given the input:

    "\"\n\n<\\r9999999Y\\\"\n9\\99999999999999g9999999999999g999999g\"\n"

the test threw an exception:

    Invalid_argument("Char.chr")
    Raised at file "stdlib.ml", line 30, characters 20-45
    Called from file "src/core/CCSexp_lex.mll", line 46, characters 30-75
[...]

PR #339, opened 2020-12-08
Init files for fuzzing

darrenldl:
Spawned from #338 

Adds the following fuzzing tests:
- `ccsexp_parse_string_does_not_crash`
  - The fuzzing test that revealed #338 
- `ccutf8_string_uchar_to_bytes_is_same_as_simple_version`
  - Tests `CCUtf8_string.uchar_to_bytes` against a simpler implementation of the equivalent

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ setup.*
 .ignore
 _opam
 *.exe
+fuzz-*-input
+fuzz-*-output
--- a/fuzz/ccsexp_parse_string_does_not_crash.ml
+++ b/fuzz/ccsexp_parse_string_does_not_crash.ml
@@ -1,0 +1,3 @@
+let () =
+  Crowbar.add_test ~name:"ccsexp_parse_string_does_not_crash" [ Crowbar.bytes ]
+    (fun s -> CCSexp.parse_string s |> ignore)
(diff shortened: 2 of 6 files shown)


c-cube:
I get this after trying to call `run.sh`:

```
[-] Hmm, looks like the target binary terminated before we could complete a
    handshake with the injected code. There are two probable explanations:

    - The current memory limit (50.0 MB) is too restrictive, causing an OOM
      fault in the dynamic linker. This can be fixed with the -m option. A
      simple way to confirm the diagnosis may be:

      ( ulimit -Sv $[49 << 10]; /path/to/fuzzed_app )

      Tip: you can use http://jwilk.net/software/recidivm to quickly
      estimate the required amount of virtual memory for the binary.

    - Less likely, there is a horrible bug in the fuzzer. If other options
      fail, poke <lcamtuf@coredump.cx> for troubleshooting tips.
```

darrenldl:
Is that with both tests? Both run fine on my system : /

darrenldl:
I've increased the virtual memory allocated to 100MiB, but I'm not confident that's the issue.

darrenldl:
Aha, I managed to replicate that when I'm back to switch `4.08.1` instead of `4.08.1+afl` - were you using a `+afl` switch when running the scripts?

c-cube:
I'm using 4.11.1+afl, but maybe there's some opam state that has not been properly updated when I upgraded my switch.

darrenldl:
Both tests still run fine for me after switching to `4.11.1+afl`, hm...

c-cube:
It was my fault, sorry. I thought I had 4.11.1+afl, but it turns out opam had rejected my `pin` (I needed `opam pin ocaml-variants 4.11.1+afl --unlock-base`) so I had a vanilla 4.11.

This seems to work, it's fuzzing right now!! :grin:

PR #342, opened 2020-12-13
CCFormat: append combinator and space/break/cut aliases

### LINKED ISSUE #314: Add chaining/sequencing combinator in CCFormat
# The problem

Currently, `CCFormat` lacks a clear combinator way to sequence two different printers.  

When printing things in boxes, the "combinators" approach provided by `CCFormat` is much nicer to use than the stdlib-compatible imperative `{open,close}_*box` API:
```ocaml
(* Format *)
pp_open_vbox fmt 2;
pp fmt a;
pp_close_box fmt ()

(* CCFormat combinators *)
vbox ~i:2
  pp
  fmt a
```

But suppose inside `...` I want to apply two printers `pp` and `pp'` in sequence.  Typically, I can imperatively sequence those two printers by writing
```
pp fmt a;
pp' fmt b
```
Clearly, this works with the imperative API (and is the usual way of chaining printers)--but in order for me to put the two printers within a box _combinator_ I have to define a new printer function:
```
vbox ~i:2
  (fun fmt () ->
    pp fmt a;
    pp' fmt b
  )
  fmt ()
```
which is not unbearable, but now quite a bit more noisy.  Another method to avoid imperative sequencing altogether is to use the `pair` combinator:
```
vbox ~i:2
  (pair ~sep:(return "") pp pp' (a, b))
  fmt ()
```
(and `triple`, `quad`, or even `list` for arbitrarily many sequence steps).  But this solution feels quite hacky and does not read that elegantly either.

# A solution from dbuenzli's [`Fmt`](https://github.com/dbuenzli/fmt)

`Fmt`'s attempt at a solution to this problem is an [`append`](https://erratique.ch/software/fmt/doc/Fmt/index.html#val-append) combinator for combining two printers.  It's defined as `append pp pp' fmt a = pp fmt a; pp' fmt a`.  It's not perfect, and it is a bit odd that both printers receive the same argument, which is not how you'd expect to do sequencing--but you can use `const` to pre-apply arguments and do something like
```
vbox ~i:2
  (append (const pp a) (const pp' b))
  fmt ()
```
While still a bit noisy (with the `const` calls), I find this much more readable, concise, and combinator-friendly.   And if you introduce infix operators `(@@@) = const` (similar to `@@` for usual function application, and for operator precedence rules) and `(<+>) = append` (`Fmt` chooses `++`, but I'm suggesting `<+>` here because it has lower operator precedence than `@`, per OCaml's [operator precedence rules](https://caml.inria.fr/pub/docs/manual-ocaml/expr.html#ss%3Aprecedence-and-associativity), and thus requires fewer parentheses):
```
vbox ~i:2
  (pp @@@ a <+> pp @@@ b)
  fmt ()
```
or maybe you're not a fan of `@@@`, in which we can just copy `Fmt`'s `++`:
```
vbox ~i:2
  (const pp a ++ const pp' b)
  fmt ()
```

# Conclusion

In any case, I personally find that having some combinator (and preferably with an infix version) for sequencing printers would be very nice; I'd like to see what you all think about this feature and its implementation details.

kwshi:
Resolves #314.  This PR adds the following to the `CCFormat` module:
- `space` (alias to `pp_print_space`), `cut` (alias to `pp_print_cut`), and `break` (wrapper around `pp_print_break` that turns it into an `(int * int) printer`).
- `append` (runs two `unit printer`s sequentially), `++` (alias to `append`), and `append_l` (runs a list of printers sequentially).

### DIFF
--- a/src/core/CCFormat.ml
+++ b/src/core/CCFormat.ml
@@ -117,7 +117,8 @@ let append ppa ppb fmt () =
   "foo" (to_string_test (append (return "foo") (return "")))
 *)
 
-let append_l = List.fold_left append (return "")
+let append_l ppl fmt () =
+  List.iter (fun pp -> pp fmt ()) ppl
 
 (*$= append_l & ~printer:(fun s -> CCFormat.sprintf "%S" s)
   "" (to_string_test @@ append_l [])
--- a/src/core/CCFormat.ml
+++ b/src/core/CCFormat.ml
@@ -107,6 +107,24 @@ let triple ?(sep=return ",@ ") ppa ppb ppc fmt (a, b, c) =
 let quad ?(sep=return ",@ ") ppa ppb ppc ppd fmt (a, b, c, d) =
   Format.fprintf fmt "%a%a%a%a%a%a%a" ppa a sep () ppb b sep () ppc c sep () ppd d
 
+let append ppa ppb fmt () =
+  ppa fmt ();
+  ppb fmt ()
+
+(*$= append & ~printer:(fun s -> CCFormat.sprintf "%S" s)
+  "foobar" (to_string_test (append (return "foo") (return "bar")))
+  "bar" (to_string_test (append (return "") (return "bar")))
+  "foo" (to_string_test (append (return "foo") (return "")))
+*)
+
+let append_l = List.fold_left append (return "")
+
+(*$= append_l & ~printer:(fun s -> CCFormat.sprintf "%S" s)
+  "" (to_string_test @@ append_l [])
+  "foobarbaz" (to_string_test @@ append_l (List.map return ["foo"; "bar"; "baz"]))
+  "3141" (to_string_test @@ append_l (List.map (const int) [3; 14; 1]))
+*)
+
 let within a b p out x =
   string out a;
   p out x;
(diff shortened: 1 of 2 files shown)


src/core/CCFormat.ml
```diff
@@ -104,6 +107,24 @@ let triple ?(sep=return ",@ ") ppa ppb ppc fmt (a, b, c) =
...
+  ppa fmt ();
+  ppb fmt ()
+
+(*$= append & ~printer:(fun s -> CCFormat.sprintf "%S" s)
+  "foobar" (to_string_test (append (return "foo") (return "bar")))
+  "bar" (to_string_test (append (return "") (return "bar")))
+  "foo" (to_string_test (append (return "foo") (return "")))
+*)
+
+let append_l = List.fold_left append (return "")
```
c-cube, line 120:
```suggestion
let append_l ppl fmt () =
  List.iter (fun pp -> pp fmt ()) ppl
```

c-cube: this would be more efficient (allocates less)****

src/core/CCFormat.mli
```diff
@@ -34,6 +34,15 @@ val exn : exn printer
 (** Printer using {!Printexc.to_string}.
     @since 3.0 *)
 
+val space : unit printer
```
c-cube, line 37: needs a docstring `@since NEXT_RELEASE` :-)

kwshi: Added in 94aa3cc

PR #349, opened 2021-01-25
feat: add CCEither module

FardaleM:


c-cube:
there needs to be a `module Either = CCEither` in containers.ml, but otherwise it looks neat! :)

c-cube:
Perfect, thank you!

PR #351, opened 2021-02-05
Small upgrade to fuzzing suite scripts

darrenldl:
This PR adds two small scripts I made for a fuzzing pipeline
- `clean.sh` just removes all fuzzing input/output files
- `run_all.sh` does parallel fuzzing of jobs in batches with (CPU core count - 1) afl instances with time out in stages
  - For instance, if we have 4 CPU core count, and we have 5 fuzzing tests, then 3 tests are run in parallel until time out, then remaining 2 tests are run in parallel until time out
  - I currently set the time out to 10m, but you can adjust that at `test_timeout="10m"` to something else

`run.sh` is also upgraded mostly for interacting with `run_all.sh`

Note: CPU core count - 1 instead of exactly CPU core count afl instances is desirable because afl uses 100% of CPU, and leaving no processing time to other processes makes it very difficult to continue controlling your system, so we leave one core spare just so we can intervene etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ _opam
 *.exe
 fuzz-*-input
 fuzz-*-output
+fuzz-logs/
--- a/fuzz/clean.sh
+++ b/fuzz/clean.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+script_dir=$(dirname $(readlink -f "$0"))
+
+rm -r "$script_dir"/../fuzz-*-input
+rm -r "$script_dir"/../fuzz-*-output
--- a/fuzz/run.sh
+++ b/fuzz/run.sh
@@ -2,16 +2,20 @@
 
 script_dir=$(dirname $(readlink -f "$0"))
 
-echo "Building"
+skip_build=$2
 
-dune build @all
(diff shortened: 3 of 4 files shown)


darrenldl:
Oop, some more modifications needed, don't merge yet.

darrenldl:
Okay, should be good now. What it looks like:

```
$ ./run_all.sh
Building
Entering directory '/home/darren/ocaml-containers'
Done: 0/0 (jobs: 0)
Fuzzing tests available:
- ccsexp_csexp_reparse
- ccsexp_parse_string_does_not_crash
- ccutf8_string_uchar_to_bytes_is_same_as_simple_version

Fuzzing start time: 2021-02-05 15:37

Starting 3 tests

Running 3 tests in parallel
  Starting ccsexp_csexp_reparse
  Starting ccsexp_parse_string_does_not_crash
  Starting ccutf8_string_uchar_to_bytes_is_same_as_simple_version
Waiting for 10m
Terminating tests

0 / 3 tests remaining


Test end: 2021-02-05 15:47

Time elapsed: 10 minutes
========================================
All 3 tests passed
```

`fuzz-logs/` contains stdout log of the fuzzing tests

PR #352, opened 2021-02-07
CCMap: implement {of,add}_*_with family of function with update

aspiwack:
This is comparable in conciseness and clarity as an explicit try/with
but it paves the way for a more efficient implementation using the
`update` from the Stdlib which, I presume, uses a one-pass algorithm.

I didn't do the same for `Hashtbl` because it isn't clear that there is a
beneficial such implementation for hash tables.


c-cube:
merci bien :grin:

Issue #346, opened 2020-12-28
Integer overflow errors when compiling to JS via jsoo


darrenldl:
We need to compile containers to JS for our [doodlinator](https://github.com/daypack-dev/doodlinator) project via js_of_ocaml.

During compilation, we'd receive the following integer overflow errors (which I believe comes from jsoo compiling containers):
```
$ make
dune build bookmarklet/doodlinator.bc.js web/webpage.bc.js
 js_of_ocaml .js/containers/containers.cma.js
Warning: integer overflow: integer 0x5555555555555555 (-3074457345618258603) truncated to 0x55555555 (1431655765); the generated code might be incorrect.
Warning: integer overflow: integer 0x3333333333333333 (3689348814741910323) truncated to 0x33333333 (858993459); the generated code might be incorrect.
Warning: integer overflow: integer 0x3333333333333333 (3689348814741910323) truncated to 0x33333333 (858993459); the generated code might be incorrect.
Warning: integer overflow: integer 0xf0f0f0f0f0f0f0f (1085102592571150095) truncated to 0xf0f0f0f (252645135); the generated code might be incorrect.
```

The last integer matches `Sys.max_array_length`, which is used in `core/CCVector` (and `data/CCPersistentArray`, `data/CCPersistentHashtbl`, but we're not using `containers-data`).

#### Version info
```
ocaml: 4.08.1
containers: 3.1
jsoo: 3.8.0
```

#### System info
Linux 64 bit

c-cube:
This is likely caused by the module `CCShimsInt_.ml`, produced by `src/core/mkshims.ml`. It's produced because the compiler is in 64 bits mode.

I think we need to use dune configurator to obtain ocamlc information for the _target_ architecture, not the _current_ architecture.

c-cube:
Can you please try the branch fix-346 ?

darrenldl:
Still gives the same error after `opam pin add containers https://github.com/c-cube/ocaml-containers.git#fix-346` - did i get the command wrong?

c-cube:
Seems like the right command, did it reinstall everything?

darrenldl:
Yep, and I reinstalled everything again as well.

c-cube:
In `_build/log`, is there any mention of `int_size`? or "target word size" ?

darrenldl:
Snippet of `_build/log`
```
#  ; ocaml_config =
#      { version = "4.08.1"
...
#      ; architecture = "amd64"
[...]

PR #358, opened 2021-02-21
Fix integer overflow warning on jsoo (#346)

copy:
- Remove popcount from shims
- Express large integer literals using bitshifts

Also ran `make test` in a 4.11.1+32bit switch since it doesn't seem to be covered by CI at the moment.


c-cube:
thanks! :)

PR #362, opened 2021-03-29
CCSeq: Add `for_all` and `exists`

### LINKED ISSUE #361: Add `CCSeq.exists` and `CCSeq.for_all`
Would you be interested in adding `exists` and `for_all` to `CCSeq` or are they considered outside of the scope of this package and better covered by `oseq`?

I'm asking because I find them useful and because `CCSeq` already has `exists2` and `for_all2`.

bbc2:
The functions are implemented the same way as in `oseq` and their documentation is inspired from their counterparts in `Stdlib.List`.

Closes #361.


c-cube:
Excellent! Even the @since annotation, well done and thank you :-)

PR #365, opened 2021-04-15
feat(CCRAL): add `get_and_remove_exn` operation

jberdine:
It seems like a waste to repeat the search for an index to both get
the element at that index and then to remove it. The added
`get_and_remove_exn` operation performs a `remove` but returns the
found element rather than forgetting it.

### DIFF
--- a/src/data/CCRAL.ml
+++ b/src/data/CCRAL.ml
@@ -125,6 +125,17 @@ let remove l i = _remove [] l i
   [1;2;4] (to_list @@ remove (of_list [1;2;3;4]) 2)
 *)
 
+let rec _get_and_remove_exn found prefix l i =
+  let x, l' = front_exn l in
+  if i=0
+  then (found := Some x; List.fold_left (fun l x -> cons x l) l' prefix)
+  else _get_and_remove_exn found (x::prefix) l' (i-1)
+
+let get_and_remove_exn l i =
+  let found = ref None in
+  let l' = _get_and_remove_exn found [] l i in
+  (Option.get !found, l')
+
 let rec _map_tree f t = match t with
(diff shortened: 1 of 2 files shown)


src/data/CCRAL.ml
```diff
@@ -116,11 +116,22 @@ let front_exn l = match l with
...
+let rec _get_and_remove_exn found prefix l i =
+  let x, l' = front_exn l in
+  if i=0
+  then (found := Some x; List.fold_left (fun l x -> cons x l) l' prefix)
+  else _get_and_remove_exn found (x::prefix) l' (i-1)
+
+let get_and_remove_exn l i =
+  let found = ref None in
+  let l' = _get_and_remove_exn found [] l i in
+  (Option.get !found, l')
```
c-cube, line 133:
```suggestion
let rec _get_and_remove_exn prefix l i =
  match front l with
  | None -> None, l
  | Some (x,l') ->
    if i=0
    then (Some x, List.fold_left (fun l x -> cons x l) l' prefix)
    else _get_and_remove_exn found (x::prefix) l' (i-1)

let get_and_remove_exn l i =
  _get_and_remove_exn found [] l i
```

this would be more direct and would avoid the ref (just return a tuple directly), I believe.

c-cube (changes requested):
Thank you for the contribution! I have a few comments. First, this needs to be rebased :-), and I think it could be simplified. Please also add a test.

src/data/CCRAL.ml
jberdine: Yes, I'm guilty of premature optimization by default here.

c-cube:
Thank you! :)

PR #374, opened 2021-05-18
fix(hash): resolve ambiguous doc comment of bytes

favonia:
There's Warning 50 now. :sweat_smile:


c-cube:
good catch, thank you :joy:

PR #376, opened 2021-05-22
use bytes instead of int array for CCBV

FardaleM:


src/data/CCBV.ml
```diff
@@ -35,31 +50,31 @@ let count_bits_ n =
...
+    let b = if default
+      then Bytes.make n all_ones_
+      else Bytes.make n zero
     in
     (* adjust last bits *)
-    let r = size mod width_ in
+    let r = mod_ size in
     if default && r <> 0 then (
-      Array.unsafe_set a (n-1) lsb_masks_.(r);
+      Bytes.unsafe_set b (n-1) (Bytes.unsafe_get lsb_masks_ r);
```
c-cube, line 75:
```suggestion
      Bytes.unsafe_set b (n-1) ((1 lsl r) - 1);
```

this should set `r` bits to one.

src/data/CCBV.ml
```diff
@@ -647,4 +662,4 @@ let pp out bv =
   "bv {00001}" (CCFormat.to_string pp (of_list [4]))
 *)
 
-let __to_word_l bv = Array.to_list bv.a
+let __to_word_l bv = CCString.to_list (Bytes.unsafe_to_string bv.b)
```
c-cube, line 665:
```suggestion
let __to_word_l bv =
  let l = ref [] in
  Bytes.iter (fun c -> l := c :: !l) bv.b;
  List.rev !l
```

it's a bit sad to pull a dep on this module just for this one function :grin:

c-cube (changes requested):
Excellent :woman_dancing: . Just a few changes but I think this will be super helpful.

src/data/CCBV.ml
```diff
@@ -8,20 +8,35 @@
...
 
-(** We use OCamls ints to store the bits. We index them from the
+(* Helper functions *)
+let[@inline] get_ b i = Char.code (Bytes.get b i)
+let[@inline] unsafe_get_ b i = Char.code (Bytes.unsafe_get b i)
+
+let[@inline] set_ b i v = Bytes.set b i (Char.unsafe_chr v)
+let[@inline] unsafe_set_ b i v = Bytes.unsafe_set b i (Char.unsafe_chr v)
+
+let[@inline] mod_ n = (n lsl (Sys.word_size - 4)) lsr (Sys.word_size - 4)
```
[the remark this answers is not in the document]
favonia: I think OCaml (at least `ocamlopt`) is smart enough to optimize away `n * 8` and `n mod 8`. If so, maybe we do not need to sacrifice readability.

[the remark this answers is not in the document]
c-cube: I think the bit shifts are readable, and they will be fast even in bytecode. Other portions of the code need bit twiddling anyway so we could be uniform.

src/data/CCBV.ml
```diff
@@ -138,19 +152,19 @@ let resize bv size =
 
 let is_empty bv =
   try
-    for i = 0 to Array.length bv.a - 1 do
-      if bv.a.(i) <> 0 then raise Exit       (* MSB of last element are all 0 *)
+    for i = 0 to Bytes.length bv.b - 1 do
+      if get_ bv.b i <> 0 then raise Exit
```
favonia, line 156: `unsafe_get_` should be okay?

src/data/CCBV.ml
```diff
@@ -1,65 +1,79 @@
...
+    let b = if default
+      then Bytes.make n all_ones_
+      else Bytes.make n zero
     in
     (* adjust last bits *)
-    let r = size mod width_ in
+    let r = mod_ size in
     if default && r <> 0 then (
-      Array.unsafe_set a (n-1) lsb_masks_.(r);
+      Bytes.unsafe_set b (n-1) (Char.unsafe_chr (lsb_masks_ r));
```
favonia, line 74: I wonder if you meant to use `unsafe_set_` here.

favonia:
I wonder if it's possible to promote `__to_word_l` to be part of the official interface. It has a reasonable semantics modulo bit-endianness. I have been wishing for a function like this to implement hashing and other things. :-)

src/data/CCBV.ml
```diff
@@ -74,21 +88,21 @@ let create ~size default =
...
+let capacity bv = mul_ (Bytes.length bv.b)
 
 let cardinal bv =
   if bv.size = 0 then 0
   else (
     let n = ref 0 in
-    for i = 0 to Array.length bv.a - 1 do
-      n := !n + count_bits_ bv.a.(i) (* MSB of last element are all 0 *)
+    for i = 0 to Bytes.length bv.b - 1 do
+      n := !n + count_bits_ (get_ bv.b i) (* MSB of last element are all 0 *)
```
favonia, line 105: Same here (`unsafe_get_`).

favonia:
PS: I am not sure how "unsafe" we want to be, but there are also `Bytes.unsafe_blit` and `Bytes.unsafe_fill` to exploit.

c-cube:
> I wonder if it's possible to promote `__to_word_l` to be part of the official interface. With bytes, it has a reasonable semantics modulo bit-endianness. I have been wishing for a function like this to implement hashing and efficient serialization. :-)

The interplay with `size` (or lack thereof, if we move in that direction indeed) seems non-trivial to me, at least for hashing. For serialization I think it's fine if there's also a reverse function. Maybe in a submodule dedicated to conversion with lists/seqs of bytes?

Btw @favonia do you happen to be on IRC? sometimes it's nice to talk live :-) (on #ocaml on libera.chat).

favonia:
> The interplay with `size` (or lack thereof, if we move in that direction indeed) seems non-trivial to me, at least for hashing. For serialization I think it's fine if there's also a reverse function. Maybe in a submodule dedicated to conversion with lists/seqs of bytes?

Got it. Perhaps all of these belong to another PR, then.

c-cube:
> Got it. Perhaps all of these belong to another PR, then.

They do, please keep them coming :slightly_smiling_face:

copy:
I believe most functions in this would be faster if they operated on 64 bits at a time (or 32 bits for 32-bit platforms and jsoo). Unsafe operations are available as primitives (`external get_int{32,64}_unsafe : string -> int -> int{32,64} = "%caml_bytes_get{32,64}u"`).

c-cube:
@copy isn't there some overhead to using these externals? Are they unboxed? I'd want to see benchmarks first :slightly_smiling_face: .

Also they're >= 4.08 only.

copy:
> @copy isn't there some overhead to using these externals? Are they unboxed? I'd want to see benchmarks first.

Yes, they are unboxed (even without flambda): https://godbolt.org/z/z4vMjK6h4

As for performance, on 4.12.0+flambda:

```
┌──────────┬──────────┐
│ Name     │ Time/Run │
├──────────┼──────────┤
│ negate8  │  43.83ns │
│ negate64 │   7.46ns │
└──────────┴──────────┘
```

```ocaml
let[@inline] unsafe_get_ b i = Char.code (Bytes.unsafe_get b i)
let[@inline] unsafe_set_ b i v = Bytes.unsafe_set b i (Char.unsafe_chr v)

let negate8 b =
  for i = 0 to Bytes.length b - 1 do
    unsafe_set_ b i (lnot (unsafe_get_ b i))
  done

external get_int64_unsafe : bytes -> int -> int64 = "%caml_bytes_get64u"
external set_int64_unsafe : bytes -> int -> int64 -> unit = "%caml_bytes_set64u"

let negate64 b =
    for i = 0 to Bytes.length b / 8 - 1 do
        let i = i * 8 in
        set_int64_unsafe b i (Int64.lognot (get_int64_unsafe b i))
    done

open Core_bench
let () =
  let xs = Bytes.create 64 in
  Bench.bench [
    Bench.Test.create ~name:"negate8" (fun () ->
        negate8 xs
      );
    Bench.Test.create ~name:"negate64" (fun () ->
        negate64 xs
      );
  ]
```

Gbury:
As far as I can tell, the `Bytes.get_X ` functions are actually compiler primitive (e.g. `caml_bytes_get32`) and are compiled to loads from memory (with or without bound checks depending on whether it is `unsafe`), so there should be no real overhead, ^^

c-cube:
Sure, but since they're not aligned it's probably going to do a shift+`lor` to reconstruct the integer in the first place? Not sure why that should be faster.

Gbury:
> Sure, but since they're not aligned it's probably going to do a shift+`lor` to reconstruct the integer in the first place? Not sure why that should be faster.

That depends on whether the architecture supports unaligned (as in, aligned on words) loads or not: if the arch supports unaligned loads, then it is a single load, else, there is some reconstruction involved indeed.
For reference  `amd64` (e.g. `x86_64`) allows unaligned loads.

c-cube:
Ok, I stand corrected. @copy you need to process the last (up to) 7 bytes in your benchmark though :)

Gbury:
Small note: In the godbolt link, most of the instructions of the `set_` function actually seem to be from the bound check (you can see the `movq    -8(%rax), %rsi` reading the header of the bytes passed as argument, and then the computation of the actual bytes length in number of bytes that goes on basically until the jump). Not sure that it actually affects performances as processors nowadays are pretty good at guessing that bounds check do not raise most of the time, ^^

c-cube:
I'm pretty impressed by this, @copy . Thanks for the suggestion, it could be worth it. I don't think we have a way, sadly, to special case jsoo (as it's compiled for 64 bits bytecode).

This would bump the minimal version of OCaml to 4.08. Maybe it's time to finally do that; these `bytes` operators have been in my sights for a while :slightly_smiling_face: .

copy:
> Small note: In the godbolt link, most of the instructions of the set_ function actually seem to be from the bound check (you can see the movq -8(%rax), %rsi reading the header of the bytes passed as argument, and then the computation of the actual bytes length in number of bytes that goes on basically until the jump).

Are you sure? The `u` in `caml_bytes_get64u` stands for unsafe, I don't think it should do bounds checking. I also don't see any exceptions in the code.

> This would bump the minimal version of OCaml to 4.08. Maybe it's time to finally do that; these bytes operators have been in my sights for a while 

They were introduced in https://github.com/ocaml/ocaml/commit/4c704c9e296189c65cfc421f6fe4d3de37242de0, so 4.07.0 apparently. In older versions you could use `%caml_string_set64u`.

c-cube:
The `@since ` annotations indicate 4.08. The external didn't exist before, did it?

copy:
> The @since annotations indicate 4.08. The external didn't exist before, did it?

The external was added in 4.07 (in the commit I linked above), and previously existed under a different name.

Gbury:
> > .
> 
> Are you sure? The `u` in `caml_bytes_get64u` stands for unsafe, I don't think it should do bounds checking. I also don't see any exceptions in the code.
> 
> >

Yes you're correct, my comment was about the `set_` function defined at line 4 in your example, and indeed in the rest whenever you use the unsafe version, there are no bound checks as expected (so in a nutshell my comment was rather useless, ^^).

favonia:
Wow this is all very exciting. I wonder if we should set up some benchmark before this PR to experiment all kinds of interesting implementations. (I am very curious about whether this PR is faster than the current CCBV.)

c-cube:
What's the status on that PR, @FardaleM ? :slightly_smiling_face:

FardaleM:
I plan to add benchmarks with the different ideas but currently I don't have much time to do that.

c-cube:
closed in favor of #394, if there's benchmarks they should be added there :)

PR #378, opened 2021-05-23
feat(list): add sorted_remove and sorted_diff

### LINKED ISSUE #377: Add CCList.sorted_remove and CCList.sorted_diff?
I wonder if it makes sense to have the negative version of `sorted_insert` and `sorted_merge`: `sorted_remove` and `sorted_diff`. These two seem convenient when I want to treat lists as sets.

favonia:
Close #377.

### DIFF
--- a/src/core/CCList.ml
+++ b/src/core/CCList.ml
@@ -797,6 +797,30 @@ let sorted_insert ~cmp ?(uniq=false) x l =
       List.mem x (sorted_insert ~cmp:CCInt.compare x l))
 *)
 
+let sorted_remove ~cmp ~key l =
+  let rec aux cmp key left l = match l with
+    | [] -> List.rev left
+    | y :: tail ->
+      match cmp key y with
+        | 0 -> aux cmp key left tail
+        | n when n<0 -> List.rev_append left l
+        | _ -> aux cmp key (y::left) tail
+  in
+  aux cmp key [] l
+
+(*$Q
+    Q.(pair small_int (list small_int)) (fun (key,l) -> \
+      let l = List.sort Stdlib.compare l in \
+      is_sorted ~cmp:CCInt.compare (sorted_remove ~cmp:CCInt.compare ~key l))
+    Q.(pair small_int (list small_int)) (fun (key,l) -> \
+      let l = List.sort Stdlib.compare l in \
+      let l' = sorted_remove ~cmp:CCInt.compare ~key l in \
+      List.length l' = List.length l - count (Int.equal key) l)
+    Q.(pair small_int (list small_int)) (fun (key,l) -> \
+      let l = List.sort Stdlib.compare l in \
+      not (List.mem key (sorted_remove ~cmp:CCInt.compare ~key l)))
+*)
+
 let uniq_succ ~eq l =
   let rec f acc l = match l with
     | [] -> List.rev acc
--- a/src/core/CCList.mli
+++ b/src/core/CCList.mli
@@ -522,6 +522,11 @@ val sorted_insert : cmp:('a -> 'a -> int) -> ?uniq:bool -> 'a -> 'a list -> 'a l
       [x] is not duplicated. Default [false] ([x] will be inserted in any case).
     @since 0.17 *)
 
+val sorted_remove : cmp:('a -> 'a -> int) -> key:'a -> 'a list -> 'a list
+(** [sorted_insert ~cmp ~key l] removes [key] from a sorted list [l] such that
+    the return value is sorted too.
+    @since NEXT_RELEASE *)
+
 val uniq_succ : eq:('a -> 'a -> bool) -> 'a list -> 'a list
 (** [uniq_succ ~eq l] removes duplicate elements that occur one next to the other.
     Examples:
--- a/src/core/CCListLabels.mli
+++ b/src/core/CCListLabels.mli
@@ -525,6 +525,11 @@ val sorted_insert : cmp:(('a -> 'a -> int) [@keep_label]) -> ?uniq:bool -> 'a ->
       [x] is not duplicated. Default [false] ([x] will be inserted in any case).
     @since 0.17 *)
 
+val sorted_remove : cmp:(('a -> 'a -> int) [@keep_label]) -> key:('a [@keep_label]) -> 'a list -> 'a list
+(** [sorted_insert ~cmp ~key l] removes [key] from a sorted list [l] such that
+    the return value is sorted too.
+    @since NEXT_RELEASE *)
+
 val uniq_succ : eq:(('a -> 'a -> bool) [@keep_label]) -> 'a list -> 'a list
 (** [uniq_succ ~eq l] removes duplicate elements that occur one next to the other.
     Examples:


c-cube (changes requested):
Two things:

- for consistency with `sorted_insert` I think we should not make `key` a label, but just a positional argument
- should there be a `?all:bool` argument to control whether to remove only the first occurrence, or all of them?

src/core/CCList.mli
```diff
@@ -501,6 +501,11 @@ val sorted_merge : cmp:('a -> 'a -> int) -> 'a list -> 'a list -> 'a list
 (** [sorted_merge ~cmp l1 l2] merges elements from both sorted list using
     the given comparison function [cmp]. *)
 
+val sorted_diff : cmp:('a -> 'a -> int) -> 'a list -> 'a list -> 'a list
+(** [sorted_merge ~cmp l1 l2] returns the elements in [l1] that are not in [l2].
+    Both lists are assumed to be sorted with respect to [cmp] and
+    duplicate elements are treated individually. *)
```
c-cube, line 507:
```suggestion
val sorted_diff : cmp:('a -> 'a -> int) -> 'a list -> 'a list -> 'a list
(** [sorted_diff ~cmp l1 l2] returns the elements in [l1] that are not in [l2].
    Both lists are assumed to be sorted with respect to [cmp] and
    duplicate elements are treated individually.
    @since NEXT_RELEASE *)
```

favonia:
LOL I unintentionally tagged someone...

src/core/CCList.ml
```diff
@@ -745,6 +745,34 @@ let sorted_merge ~cmp l1 l2 =
...
 
+let sorted_diff ~cmp l1 l2 =
+  let rec recurse cmp acc l1 l2 = match l1,l2 with
+    | [], _ -> List.rev acc
+    | _, [] -> List.rev_append acc l1
+    | x1::l1', x2::l2' ->
+      let c = cmp x1 x2 in
+      if c < 0 then recurse cmp (x1::acc) l1' l2
+      else if c > 0 then recurse cmp acc l1 l2'
+      else recurse cmp acc l1' l2'
```
c-cube, line 756:
```suggestion
      else recurse cmp acc l1' l2
```

if there are several consecutive occurrences of `x1` in `l1` but only one in `l2`, correct?

favonia: This is intentional---otherwise it would be wrong for multisets. The idea is that `sorted_diff (sorted_merge l1 l2) l2 = l1`. Actually, maybe this should be added to the quickchecking...

favonia: For my applications it actually does not matter, so I am not strongly against the other behavior.

favonia: Actually, I am not sure if the other behavior is very useful. If the elements are unique in individual lists, then `sorted_diff` and `sorted_merge_unique` will handle the lists correctly. If the elements are not unique (lists as multisets), then `sorted_diff` and `sorted_merge` will handle them correctly. In either case, the current code seems to work well, and thus I prefer keeping it.

c-cube: My point is, what if you use the sorted list as a multiset and want to remove only one element, not all its occurrences? :)

favonia: Could you possibly give an example? This thread is for `sorted_diff`, and I felt `sorted_diff` is already doing exactly what you specify. As for `sorted_remove`, it has been adjusted.

c-cube: say with this test:

```ocaml
(*$=
  [1;3] (sorted_diff [1;2;2;3] [2])
*)
```

favonia: I feel for multisets, it might make more sense to have `{a,a,a,a,b,b} \ {a,a,b} = {a,a,b}`, governed by the formula
```
multiplicity(x\y)(a) = max(0, multiplicity(x)(a) - multiplicity(y)(a))
```
Currently, `sorted_merge l [x]` is the same as `sorted_insert x l`, and for consistency maybe `sorted_diff l [x]` should be the same as `sorted_remove x l`, removing only one element from `l`?

PS: Currently I am not using multisets so none of these would affect me.

c-cube: I agree on the laws for `merge l [x] = insert x l` and the symmetric, but that's also the point: there's already some arguments to specify whether insert can produce duplicates or not, and similarly delete should be able to remove all elements or only one. The default value of the optional argument to control that should be the one that respects the laws.

favonia: I am sorry but I am still confused. Perhaps you missed the changes to `sorted_remove` which added the optional argument `all` with the default value `false`? By default, `sorted_remove` will remove only one occurrence. If this is different from your impression, I encourage you to check https://github.com/c-cube/ocaml-containers/pull/378/commits/6d2dc4ccf44628a5453ff8a79c677c1e01c9802e which changed the behavior of `sorted_remove`.

One of the following must be false according to your design, but I am not sure which one.

1. `[1;3] = diff [1;2;2;3] [2]` (this is NOT satisfied by the current code)
2. `remove 2 [1;2;2;3] = diff [1;2;2;3] [2]` (this is satisfied by the current code)
3. `[1;2;3] = remove 2 [1;2;2;3]` (this is also satisfied by the current code)

favonia: (I realized my commit messages did not explain the changes. Sorry about that...)

c-cube: I'm saying there's a `?all:bool` argument to remove that makes 1. true by default, but when set to false makes 3. true. Is that what you did? I'm having a hard time following both commits and discussions, sorry. It might be already what the code does.

favonia: Current (after https://github.com/c-cube/ocaml-containers/commit/6d2dc4ccf44628a5453ff8a79c677c1e01c9802e) `?all:bool` is `false` by default, which I thought was more consistent with the rest of the library, and my initial choice when drafting this PR seems to be suboptimal. Anyways, I chose the current (default) behaviors based on the following equations for lists as multisets:
```
diff (merge l1 l2) l2 = l1
remove x (insert x l) = l
diff l [x] = remove x l
```
The second equation will force the default value of `?all:bool` to be `false`.

c-cube: A last thing, sorry :sweat_smile: . There's `sorted_merge_uniq` which explicitly targets sets (removes duplicates). I suppose the pendant of that would be `sorted_diff_uniq` (?!) which makes sure to remove all occurrences of common elements?

favonia: I have to think about it... what are the equations `sorted_diff_uniq` should satisfy?

src/core/CCList.mli
```diff
@@ -510,6 +525,16 @@ val sorted_merge_uniq : cmp:('a -> 'a -> int) -> 'a list -> 'a list -> 'a list
     removes duplicates.
     @since 0.10 *)
 
+val sorted_diff_uniq : cmp:('a -> 'a -> int) -> 'a list -> 'a list -> 'a list
+(** [sorted_diff_uniq ~cmp l1 l2] collects the elements in [l1] that are not in [l2]
+    and then remove duplicates.
+    Both lists are assumed to be sorted with respect to [cmp] and
+    duplicate elements in the input lists are treated individually;
+    for example, [sorted_diff_uniq ~cmp [1;1;1;2;2] [1;2;2;2]] would be [[1]].
```
c-cube, line 533: would be `[[1;1]` I believe?

src/core/CCList.ml
```diff
@@ -866,6 +958,40 @@ let sorted_merge_uniq ~cmp l1 l2 =
...
+    | x1::l1', x2::l2' ->
+      let c = cmp x1 x2 in
+      if c < 0 then recurse ~cmp (push ~cmp acc x1) l1' l2
+      else if c > 0 then recurse ~cmp acc l1 l2'
+      else recurse ~cmp acc l1' l2'
+  in
+  recurse ~cmp [] l1 l2
+
+(*$T
+  sorted_diff_uniq ~cmp:CCInt.compare [1; 1; 1; 2; 2; 3; 5; 8; 8; 8] [1; 2; 2; 2; 2; 8; 13; 13; 13] = [1;3;5;8]
```
c-cube, line 981:
```suggestion
  sorted_diff_uniq ~cmp:CCInt.compare [1; 1; 1; 2; 2; 3; 5; 8; 8; 8] [1; 2; 2; 2; 2; 8; 13; 13; 13] = [1;3;5;8]
  sorted_diff_uniq ~cmp:CCInt.compare [1;1;1;2;2] [1;2;2;2] = [1;1]
```

src/core/CCListLabels.mli
```diff
@@ -513,6 +528,16 @@ val sorted_merge_uniq : cmp:(('a -> 'a -> int) [@keep_label]) -> 'a list -> 'a l
     removes duplicates.
     @since 0.10 *)
 
+val sorted_diff_uniq : cmp:(('a -> 'a -> int) [@keep_label]) -> 'a list -> 'a list -> 'a list
+(** [sorted_diff_uniq ~cmp l1 l2] collects the elements in [l1] that are not in [l2]
+    and then remove duplicates.
+    Both lists are assumed to be sorted with respect to [cmp] and
+    duplicate elements in the input lists are treated individually;
+    for example, [sorted_diff_uniq ~cmp [1;1;1;2;2] [1;2;2;2]] would be [[1]].
```
c-cube, line 536:
```suggestion
    for example, [sorted_diff_uniq ~cmp [1;1;1;2;2] [1;2;2;2]] would be [[1;1]].
```

c-cube:
Heh, my changes were wrong, I just reverted them on master. Thanks a lot for the patience @favonia !

PR #386, opened 2021-09-22
Rename CCOpt to CCOption and deprecate CCOpt

FardaleM:
In the stdlib the module associated with the option type is called
Option and in containers it was called CCOpt. Renaming CCOpt to CCOption
make the name of containers module uniforme with respect to the stdlib.


c-cube:
I think it's a good idea. I think `CCOpt` will remain for a long time, but as a deprecated module :).

src/core/CCOpt.mli
```diff
@@ -1,250 +1,3 @@
-(* This file is free software, part of containers. See file "license" for more details. *)
+[@@@ocaml.deprecated "use CCOption instead"]
```
c-cube, line 1:
```suggestion
(** Option module
    @deprecated use `CCOption` instead. *)
[@@@ocaml.deprecated "use CCOption instead"]
```

src/core/CCOption.mli
```diff
@@ -0,0 +1,250 @@
+(* This file is free software, part of containers. See file "license" for more details. *)
+
+(** {1 Options} *)
```
c-cube, line 3:
```suggestion
(** Options

   This module replaces `CCOpt`.
   @since NEXT_RELEASE *)
```

PR #390, opened 2021-12-12
Make CCSeq.to_array behave better with stateful sequences

shonfeder:
This PR suggests a change to `CCSeq.to_array`.

The change is motivated by the surprising behavior of `CCSeq.to_array` 
when the underlying seq depends on state. This surprising behavior can
be illustrated with this comparison between the behavior of `to_list` and
`to_array`:

```
# open Containers.Seq;;
# let r = ref 1 in 
  let s = unfold (fun i -> if i < 3 then let x = !r in incr r; Some (x, succ i) else None) 0 in
  to_array s
;;
- : int array = [|5; 6; 7|]
# let r = ref 1 in 
  let s = unfold (fun i -> if i < 3 then let x = !r in incr r; Some (x, succ i) else None) 0 in
  to_list s
;;
- : int list = [1; 2; 3]
```

The solution suggested here is to construct an intermediate list, to cash
the elements of the seq while we obtain the length of items, and thus 
prevent traversing the Seq twice. This requires the allocation of one 
intermediate list, but further allocations and traversals are avoided by
constructing the list and computing its length in one pass, and then
constructing the array from the last element, so we needn't reverse the list.

The PR also adds a test for the expected behavior with a stateful sequence.

*Some context:*

- I hit this surprising behavior doing one of this years advent of code exercises,
with a `Seq.t` reading from `stdin`.
- I discussed the issue and the possible fix a bit with @c-cube in chat.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 3.6.1
 
 - use `either` compatibility library instead of shims
+- make `CCSeq.to_array` only traverse the sequence once
 
 ## 3.6
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 ## 3.6.1
 
 - use `either` compatibility library instead of shims
-- make `CCSeq.to_array` only traverse the sequence once
 
 ## 3.6
 
--- a/src/core/CCSeq.ml
+++ b/src/core/CCSeq.ml
@@ -395,14 +395,18 @@ let of_array a =
   aux a 0
 
 let to_array l =
-  match l() with
-    | Nil -> [| |]
-    | Cons (x, _) ->
-      let n = length l in
-      let a = Array.make n x in (* need first elem to create [a] *)
-      iteri
-        (fun i x -> a.(i) <- x)
-        l;
+  (* We contruct the length and list of seq elements (in reverse) in one pass *)
+  let len, ls = fold_left (fun (i, acc) x -> (i + 1, x :: acc)) (0, []) l in
+  (* The length is used to initialize the array, and then to derive the index for
+     each item, working back from the last. This lets us only traverse the list
+     twice, instead of having to reverse it. *)
+  match ls with
+    | [] -> [||]
+    | init::rest ->
+      let a = Array.make len init in
+      (* Subtract 1 for len->index conversion and 1 for the removed [init] *)
+      let idx = len - 2 in
+      ignore (List.fold_left (fun i x -> a.(i) <- x; i - 1) idx rest);
       a
 
 (*$Q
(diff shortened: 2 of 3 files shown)


src/core/CCSeq.ml
```diff
@@ -412,6 +416,9 @@ let to_array l =
 (*$T
   of_array [| 1; 2; 3 |] |> to_list = [1;2;3]
   of_list [1;2;3] |> to_array = [| 1; 2; 3; |]
+  let r = ref 1 in \
+  let s = unfold (fun i -> if i < 3 then let x = !r in incr r; Some (x, succ i) else None) 0 in \
+  to_array s = [| 1; 2; 3; |]
```
shonfeder, line 421: I couldn't figure out a better way to do this using qtest, but happy to make any changes that would improve readability or feedback.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 3.6.1
 
 - use `either` compatibility library instead of shims
+- make `CCSeq.to_array` only traverse the sequence once
```
c-cube, line 6: that's in 3.7, but don't worry about updating the changelog, I do it at release time.

src/core/CCSeq.ml
```diff
@@ -395,14 +395,18 @@ let of_array a =
...
+  (* The length is used to initialize the array, and then to derive the index for
+     each item, working back from the last. This lets us only traverse the list
+     twice, instead of having to reverse it. *)
+  match ls with
+    | [] -> [||]
+    | init::rest ->
+      let a = Array.make len init in
+      (* Subtract 1 for len->index conversion and 1 for the removed [init] *)
+      let idx = len - 2 in
+      ignore (List.fold_left (fun i x -> a.(i) <- x; i - 1) idx rest);
```
c-cube, line 409:
```suggestion
      ignore (List.fold_left (fun i x -> a.(i) <- x; i - 1) idx rest : int);
```

I like to always type annotate `ignore` :)

src/core/CCSeq.ml
c-cube: that seems good.

c-cube (approved):
Looks pretty good to me! Thank you

CHANGELOG.md
shonfeder: I reverted this bit.

src/core/CCSeq.ml
shonfeder: Ah yeah, that is nice. Makes it a bit less spooky (I don't use `ignore` much :) ).

shonfeder:
My pleasure! Thanks for the wonderful lib :)

PR #397, opened 2022-01-25
adding opt_map to simplify result function application over optionals

ewenmaclean:
A function I use a lot when applying CCResult over optionals - e.g. in a translation of ASTs a fragment like:
```
| AST_from.Return expr -> let+ expr = opt_map trans_expr expr in AST_to.Return expr
```
where `expr` is an optional expression of type `a option` and `trans_expr` maps from `a -> ('b,'err) result`

### DIFF
--- a/src/core/CCResult.ml
+++ b/src/core/CCResult.ml
@@ -70,9 +70,11 @@ let of_exn_trace e =
 
 let opt_map f e = match e with 
   | None -> Ok None 
-  | Some x -> (match f x with 
+  | Some x ->
+    begin match f x with 
       | Ok x -> Ok (Some x)
-      | Error e -> Error e)
+      | Error e -> Error e
+    end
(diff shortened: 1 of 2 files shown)


src/core/CCResult.ml
```diff
@@ -68,6 +68,12 @@ let of_exn_trace e =
   in
   Error res
 
+let opt_map f e = match e with 
+  | None -> Ok None 
+  | Some x -> (match f x with 
+      | Ok x -> Ok (Some x)
+      | Error e -> Error e)
```
c-cube, line 75:
```suggestion
  | Some x ->
    begin match f x with 
      | Ok x -> Ok (Some x)
      | Error e -> Error e
    end
```

c-cube:
Thank you! :man_dancing:

PR #409, opened 2022-06-01
New assoc list helpers: `keys`, `values`, and `map_values`

dmbaturin:
There's a bunch of simple assoc list helpers I use quite often and I think I'm not alone, so I'd like to add them to containers:

* `keys`: returns a list of all keys.
* `values`: returns a list of all values.
* `map_values`: applies a function to all values.

If you have objections against the naming of `map_values` or any implementation details, let me know. In particular, containers lean towards making a fold over an empty list a failure, while these functions return an empty list in that case—let me know if you think it should be handled differently.


c-cube:
I don't see a fold anyway, these look pretty reasonable :)

PR #416, opened 2022-09-14
add `CCLocal_storage` to containers-thread

c-cube:
Basic local storage emulation using an atomic-protected map. The goal is for this to work both with 4.xx threads, 5.xx domains, and 5.xx threads running on a given domain.

cc @bronsa :)

### DIFF
--- a/src/threads/CCLocal_storage.mli
+++ b/src/threads/CCLocal_storage.mli
@@ -1,0 +1,46 @@
+(** Thread/Domain local storage
+
+    This allows the creation of global state that is per-domain or per-thread.
+
+    {b status} experimental
+
+    @since NEXT_RELEASE
+*)
+
+type 'a t
+
+val create : unit -> 'a t
+(** Create new storage *)
+
+val get : 'a t -> 'a option
+(** Get the content for this thread, if any. *)
+
+val get_exn : 'a t -> 'a
+(** Same as {!get}, but fails if no data was associated to this thread.
+    @raise Not_found if the data is not there. *)
+
+val get_or : default:'a -> 'a t -> 'a
+(** Same as {!get} but returns [default] if no data is associated
+    to this thread. *)
+
+val set : 'a t -> 'a -> unit
+(** Set content for this thread. *)
+
+val set_get : 'a t -> 'a -> 'a option
+(** Set content for this thread, and return the old value. *)
+
+val remove : 'a t -> unit
+(** Remove value *)
+
+val n_entries : _ t -> int
+(** Number of entries in the map currently.
+
+    Be aware that some threads might
+    have exited without cleaning up behind them. See {!with_} for
+    scope-protected modification of the variable that will cleanup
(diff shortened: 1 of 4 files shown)


c-cube:
A big design question is whether to pick:

- never empty:
  ```ocaml
  type 'a t

  val create : init:(unit -> 'a) -> unit -> 'a t

  val get : 'a t -> 'a
  ```

- can be empty (the current):
  ```ocaml
  type 'a t

  val create : unit -> 'a t

  val get : 'a t -> 'a option
  ```

Bronsa:
> A big design question is whether to pick:
> 
>     * never empty:
>       ```ocaml
>       type 'a t
>       
>       val create : init:(unit -> 'a) -> unit -> 'a t
>       
>       val get : 'a t -> 'a
>       ```
> 
>     * can be empty (the current):
>       ```ocaml
>       type 'a t
>       
>       val create : unit -> 'a t
>       
>       val get : 'a t -> 'a option
>       ```

Why not both? 
```ocaml
type 'a t

val create : init:(unit -> 'a) -> unit -> 'a t

val create_empty : unit -> 'a option t

val get : 'a t -> 'a

```

bluddy:
If deleting the thread's local storage is not part of the API, I think it makes sense to have it always contain a value.

On the other hand, what would the behavior be when you create a new thread and pass the handle to it? Can you now `get` on the TLS and expect a value? The init function can only be called when you `get`. Should it be called also on a `set` from the new thread?

src/threads/CCLocal_storage.mli
```diff
@@ -0,0 +1,43 @@
...
+
+    Be aware that some threads might
+    have exited without cleaning up behind them. See {!with_} for
+    scope-protected modification of the variable that will cleanup
+    properly (like {!Fun.protect}).
+*)
+
+val with_ : 'a t -> 'a -> (unit -> 'b) -> 'b
+(** [with_ var x f] sets [var] to [x] for this thread, calls [f()], and
+    then restores the old value of [var] for this thread. *)
```
bluddy, line 43: This seems like an odd `with_` function to me. Is it really useful?

c-cube:
@bluddy if it's always full, which makes sense, then: `get v` would initialize `v` if it was not already initialized for this thread; `set v` would just directly write the new value. You'd never observe an absence of values.

bluddy:
@c-cube I'm just concerned about possible mutable state issues. In a regular array, `init` happens at once and is a lot more straightforward. In our case here, it happens only when you observe it, and doesn't happen at all if you don't observe it.

c-cube:
Yeah I mean, if you do side-effectful stuff in a TLS initializer and then complain, it's kind of your fault :D. Unlike `Array.init`, the number of instances here is not fixed, and new threads might be started later, so initialization _must_ be lazy.

c-cube:
There's `thread-local-storage` already, and maybe it'll also land in the stdlib.

PR #417, opened 2022-09-17
add condition over os_type to preproc

FardaleM:
I started working on CCFilename, a lot of function depend on OS specific definition. Therefore I add the possibility to match the os_type using the preprocessor.


c-cube:
Seems like a good idea! Have you tested it already? :)

FardaleM:
I just tried it and it currently does not work. The fix is small but I want to improve a bit more stuff as I work with it. So I will keep it as a draft for now except if you want to have it merge early.

PR #422, opened 2022-12-19
Allow tests to run on a bytecode-only platform

shym:
As a followup to #421, this PR runs the test suite also on a bytecode-only switch, after fixing what must be fixed so that it works.


FardaleM:
Looks good, thank you!

PR #424, opened 2023-01-06
fix(CCVector): always obtain a copy of array before using unsafe_{set,get}

### LINKED ISSUE #423: segfault using CCVector with iterator invalidation
```ocaml
let vec = CCVector.of_list [1; 2; 3]
let () =
  CCVector.fold (fun () v ->
    print_int v;
    if v = 2 then begin
      CCVector.clear vec;
      CCVector.shrink_to_fit vec;
   end
  ) () vec
```

On my machine, this program prints a lot of numbers and quickly segfaults.
I think that the CCVector implementation should be *very* careful about its use of `Array.unsafe_{get,set}` functions.

c-cube:
close #423

### DIFF
--- a/src/core/CCVector.ml
+++ b/src/core/CCVector.ml
@@ -132,9 +132,11 @@ let resize_with v f size =
     v.size <- size
   ) else (
     ensure_assuming_not_empty_ v size;
-    for i = v.size to size - 1 do
-      Array.unsafe_set v.vec i (f i)
+    let {size=cur_size; vec} = v in
+    for i = cur_size to size - 1 do
+      Array.unsafe_set vec i (f i)
     done;
+    assert (size <= Array.length v.vec);
     v.size <- size
   )
 
@@ -146,6 +148,7 @@ let resize_with_init v ~init size =
     v.size <- size
   ) else (
     ensure_assuming_not_empty_ v size;
+    (* nothing will change [v] *)
     for i = v.size to size - 1 do
       Array.unsafe_set v.vec i init
     done;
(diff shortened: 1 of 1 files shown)


gasche:
I think this is a good fix.

It could still break user code in the case of people that were doing something risky that was working well before:

1. Most iterators you modified would precompute the length of the traversal at the beginning.  So they would observe mutations to not-yet-traversed elements, ignore new elements added , observe garbage for elements removed, and segfault if the backing array also shrinks. The new versions are safer but they behave different for users that only grow the array during traversal; before they would get an easy-to-explain behavior where they would observe mutation to elements within not-yet-traversed range, and not observe new elements. Now they will observe some mutations in the not-yet-traversed range, but not necessarily all of them -- they will stop observing mutations after the first resize, which will go to the new backing array. (Iterators that you wrote *not* using unsafe_get, but rather `v.vec.(i)`, those will keep observing updates even after resizes.)

2. `fold` was slightly special in that it would recompute `v.size` on each iteration. In particular, if you never remove elements but only add elements, then the previous version of `fold` will iterate on these elements -- it will keep iterating as long as new elements get added. Users might have relied on this behavior, despite the fact that it is inconsistent with the behavior of the other iterators.

To summarize:
- Right now the module is unsound in presence of iterator invalidation.
- With your fix it becomes sound, but some functions change behaviors, and the new behavior across all iterators are not consistent.
- I am not saying that you should specify carefully what happens on iterator invalidation and that it should be consistent across all functions; it is also reasonable to just ask users to not do that. But you should prepared for the fact that some users would be relying on some aspects of the previous behavior (it's not specified but they tested it to work), and their code could break.

gasche:
I think that `resize_with`, `resize_with_init` also need fixing.

PR #430, opened 2023-05-26
[doc] CCIO.File.walk

bclement-ocp:
This PR clarifies the behavior of `walk` and `read_dir ~recurse:true` by documenting that:

 - `walk p` always includes `p`
 - `read_dir ~recurse:true` only includes file paths (whereas `read_dir ~recurse:false` can include directory paths)


c-cube:
Thank you :)

PR #436, opened 2023-07-26
408 cleanup

FardaleM:
Removing function that are always present in 4.08.

### DIFF
--- a/src/core/CCFormat.ml
+++ b/src/core/CCFormat.ml
@@ -32,7 +32,6 @@ let newline = Format.pp_force_newline
 let substring out (s, i, len) : unit = string out (String.sub s i len)
 let text = Format.pp_print_text
 let option = Format.pp_print_option
-let opt = option
 let result = Format.pp_print_result
 
 let string_lines out (s : string) : unit =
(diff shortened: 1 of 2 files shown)


c-cube:
looks nice so far :)

c-cube:
I really don't think we should break `compose` :sweat_smile:

Maybe we can deprecate it, add a new name for it that doesn't clash, and we remove it in 4.00.

FardaleM:
Here is a list of things that need to be decided one:
- `ListLabel.init` takes the length as a named argument, do we want to to do the same
- `Fun.compose` have inverted argument, should we change to do the same
- `ListLabel.compare` and `ListLabel.equal` take the function on element as named arguments, should we change to do the same
- `List.partition_map` correspond to `CCList.partition_map_either` and not to the current `CCList.partition_map`, should we change to match the stdlib?
- `List.assoc` does not take the equality function as argument, we could change the label argument with an optional one to match the stdlib
- `List.sort_uniq` does not take the `cmp` function as labeled arguments

src/core/CCFormat.ml
```diff
@@ -31,6 +31,9 @@ let break fmt (m, n) = Format.pp_print_break fmt m n
 let newline = Format.pp_force_newline
 let substring out (s, i, len) : unit = string out (String.sub s i len)
 let text = Format.pp_print_text
+let option = Format.pp_print_option
+let opt = option
```
c-cube: I think we should preserve `opt` as is. I don't like that `Some x` is printed just like `x`, and I'd rather not break  that particular case. People can use `option` for the new behavior, and we need to document the difference in docstrings :)

src/core/CCFun.ml
```diff
@@ -70,7 +70,7 @@ let[@inline] with_return (type ret) f : ret =
 
 module Infix = struct
   (* default implem for some operators *)
-  let ( %> ) = compose
+  let ( %> ) f g = compose g f
```
c-cube: why??

c-cube:
ugh, let's not delay further about the compose function.

PR #442, opened 2023-11-26
CCBool: Add functions if_then and if_then_else

BridgeTheMasterBuilder:
Just some cute boolean combinators.

Example usage:

```ocaml
[true;false;true;false;true] |> List.filter_map (Bool.if_then (Fun.const 1));;
- : int list = [1; 1; 1]

[true;false;true;false;true] |> List.map (Bool.if_then_else (Fun.const 1) (Fun.const 0));;
- : int list = [1; 0; 1; 0; 1]

let t () = print_endline "Yes" in
let f () = print_endline "No" in
let select = Bool.if_then_else t f in
select true;
select false;;
Yes
No
- : unit = ()
```

etc.

### DIFF
--- a/src/core/CCBool.ml
+++ b/src/core/CCBool.ml
@@ -1,11 +1,21 @@
 (* This file is free software, part of containers. See file "license" for more details. *)
 
-
-
 type t = bool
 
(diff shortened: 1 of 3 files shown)


c-cube:
Thank you :)

PR #445, opened 2023-12-05
CCList: add unfold

### LINKED ISSUE #444: CCList.unfold
Hi Simon! 😄 

Wondering if there is a reason why `CCList` doesn't have an unfold function like e.g.:
```ocaml
let rec unfold f seed =
  match f seed with
    | None -> []
    | Some (v, next) -> v :: unfold f next
```
This has type:
`('a -> ('b * 'a) option) -> 'a -> 'b list`

One obvious problem may be that this is *unsafe* in the sense that the following would result in a stack overflow:
```ocaml
let f = fun _ -> Some (true, true);;
unfold f 0;;
(*Stack overflow during evaluation (looping recursion?).*)
```

Given that, how about having a *safe* version of unfold which also takes a `max` parameter which stops the computation if the max is reached? Or maybe having an *unsafe* unfold is fine. Anyways, if there is interest in adding this in and no qualms, I'll write up a PR this weekend! Thanks

benbellick:
Closes #444 

Let me know if the test is sufficient 😄

### DIFF
--- a/src/core/CCList.ml
+++ b/src/core/CCList.ml
@@ -276,6 +276,11 @@ let fold_flat_map_i f acc l =
   in
   aux f acc 0 [] l
 
+let rec unfold f seed =
+  match f seed with
+  | None -> []
+  | Some (v, next) -> v :: unfold f next
+
 [@@@iflt 5.1]
 
 (* keep this because it's tailrec for < 5.1 *)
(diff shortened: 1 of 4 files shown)


benbellick:
Also, how does the compiler figure out an implementation for all of the functions in `CCListLabels`? I'm surprised to see that `CCListLabels.ml` is just an import of `CCList`, even though the signatures of functions in the two _can_ differ due to the labels. I assume it must be the order?

c-cube:
Looks  good!

for labels, you can find the `-nolabels` flag in `dune` that is used to tell the typechecker to ignore labels :).

PR #447, opened 2024-01-04
CCVector: fix two labels-omitted warnings

gasche:
These warnings are silenced by the use of -nolabels in the compilation flags (which I understand is designed to make the life of the CC*Labels file easier, not let minor labeling mistakes sleep in.)


c-cube:
Merci :grin:

PR #448, opened 2024-01-06
containers.pvec

c-cube:
persistent vectors based on clojure's, in a new containers.pvec sub-library


c-cube:
TODO: optimize `append` `and `append_iter` so they operate by chunk of 16 items

c-cube:
I don't have time for a fast `append` right now, merging.

PR #449, opened 2024-01-11
CCVector: Add function foldi

BridgeTheMasterBuilder:
Just adding a version of `fold` with indices for `CCVector`.


c-cube:
Thank you!

PR #453, opened 2024-06-03
perf: use a monomorphic impl for CCMonomorphic.{min,max}

### LINKED ISSUE #452: Monomorphic.min function is slow
I realised that the min function from the Monomorphic module is using the polymorphic compare and is therefore slower than it should be.

c-cube:
close #452

### DIFF
--- a/src/monomorphic/CCMonomorphic.ml
+++ b/src/monomorphic/CCMonomorphic.ml
@@ -7,8 +7,19 @@ let ( > ) : int -> int -> bool = Stdlib.( > )
 let ( <= ) : int -> int -> bool = Stdlib.( <= )
 let ( >= ) : int -> int -> bool = Stdlib.( >= )
 let compare : int -> int -> int = Stdlib.compare
+
+[@@@ifge 4.13]
+
+let min : int -> int -> int = Int.min
+let max : int -> int -> int = Int.max
+
+[@@@else_]
+
 let min : int -> int -> int = Stdlib.min
 let max : int -> int -> int = Stdlib.max
+
+[@@@endif]
+
 let ( =. ) : float -> float -> bool = Stdlib.( = )
 let ( <>. ) : float -> float -> bool = Stdlib.( <> )
 let ( <. ) : float -> float -> bool = Stdlib.( < )


FardaleM:
From my testing, it seems enough to define `min` as:
```ocaml
let min (x:int) (y:int) = if x <= y then x else y
```
But my testing was with flambda enabled. Something that might be enough also is to do:
```ocaml
let min (x:int) (y:int) = Stdlib.min x y
```
But I did not try it to see if this is enough.
With both solution we would not need the conditional compilation.

c-cube:
```
let min (x:int) (y:int) = Stdlib.min x y
```
I've never seen before this be faster than `let min = Stdlib.min`? :/
if it is, we have a problem cause this is also how `compare`, etc are done. But anyway, I'd rather delegate to `Int.min` since it might become an intrinsic in the future.

FardaleM:
I would need to look how it is compiled. The question is, does the compiler specialised the compare function or not. If not, then it goes through the polymorphic compare which is slow. That's how I found it, the min function was calling the runtime C function on my case. We would need to check if this is the case also for the other functions. But I don't know how to check that.

c-cube:
Easiest is godbolt (https://ocaml.godbolt.org/), e.g.

https://ocaml.godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:ocaml,selection:(endColumn:58,endLineNumber:4,positionColumn:58,positionLineNumber:4,selectionStartColumn:58,selectionStartLineNumber:4,startColumn:58,startLineNumber:4),source:'%0Alet+cint+:+int+-%3E+int+-%3E+int+%3D+Stdlib.compare%0Alet+()+%3D%0A++Printf.printf+%22%25d%5Cn%22+(Sys.opaque_identity+(cint+42+41))'),l:'5',n:'1',o:'OCaml+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:ocaml5000flambda,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:ocaml,libs:!(),options:'',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+ocamlopt+5.0.0-flambda+(Editor+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

-- 
Simon Cruanes

FardaleM:
I am not sure exactly how it works and how to check this. But the question is whether or not the compiler has specialised the comparison function. If not, then I could see the call to the polymorphic compare. But just rewriting the `min` function and giving types to the argument was enough.

FardaleM:
I tried to play to see the case for `min`, but I don't see the function, so I don't know how to interpret this.

c-cube:
it might be because polymorphic `min` isn't in the compiled code, it's part of the runtime.

PR #458, opened 2024-08-24
predicate combinators: and_p and or_p

benbellick:


c-cube:
should these perhaps be called `compose_and` and `compose_or`, as they're specialized version of `compose_binop`?

benbellick:
Unless I am not understanding, it doesn't seem like a specialized version of `compose_binop`, as in `compose_binop`, the inside is the same function `f` applied to two terms `x` and `y`, whereas in `and_p` and `or_p`, the inside is two different functions `f` and `g` being applied to the same term `x`.

c-cube:
Fair enough, there could be a `compose_binop_maxi f g1 g2 x y = f (g1 x) (g2 y)` that is actually a generalization of both. But my bad :). I'd still like a more descriptive name?

benbellick:
The `p` I meant to evoke `predicate`, so how about `and_pred` or `and_predicate`?

c-cube:
Sure, let's go with `and_pred` :)

benbellick:
Done!

c-cube:
thanks! :)

PR #459, opened 2024-11-14
Add `split_result`

seprov:
# What

This PR adds `split_result` to `CCList`.

# Why

So I can later open a PR against [Lwt](https://github.com/ocsigen/lwt) to do this:
```ocaml
let lwt_collect x =
  x |> Lwt.all
  |> Lwt.map (fun r ->
         let oks, errors = CCList.split_result r in
         match errors with [] -> Ok oks | _ -> Error errors)
```

### DIFF
--- a/src/core/CCList.ml
+++ b/src/core/CCList.ml
@@ -1048,6 +1048,13 @@ let all_ok l =
     | None -> assert false
     | Some e -> Error e)
 
+let split_result results =
+  results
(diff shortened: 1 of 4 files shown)


benbellick:
I like this change 👍

c-cube:
I think I'd rather name it `CCList.split_result`, what do you think?

Issue #463, opened 2024-12-17
CCString.take_while


ygrek:
I was surprised there is drop_while but no take_while :thinking:

mobotsar:
Since the changes I made to address this were recently merged (thanks @c-cube!), I think this issue can be closed.

PR #465, opened 2025-01-01
Add `take_while` and `rtake_while` to `CCString`, addressing #463

mobotsar:
If you're open to it, I implemented a couple of apparently-missing functions in the `CCString` module, added corresponding tests, and updated `CCString.mli` and `CCStringLabels.mli` accordingly.


c-cube:
Thank you, that's useful!

Issue #471, opened 2025-01-29
CCArray.of_iter missing in containers v3.15


illusory0x0:

c-cube:
Right, it's never been there. It is in the `iter` package. Implementing reasonably efficiently requires a dynamic array  and I didn't have that at the time.

illusory0x0:
I think we can use CCVector for that.

illusory0x0:
I run the benchmark, Constructing arrays with CCVectors is a bad idea.

```
*** Run benchmarks for path "iter.to_array.1000"

Throughputs for "iter.to_array", "gen.to_array", "oseq.to_array", "ccarray.to_array" each running 3 times for at least 3 CPU seconds:
   iter.to_array:  3.16 WALL ( 3.16 usr +  0.00 sys =  3.16 CPU) @ 27720.03/s (n=87481)
                   3.16 WALL ( 3.15 usr +  0.01 sys =  3.16 CPU) @ 27671.08/s (n=87481)
                   3.16 WALL ( 3.16 usr +  0.00 sys =  3.16 CPU) @ 27667.08/s (n=87481)
    gen.to_array:  3.15 WALL ( 3.15 usr +  0.00 sys =  3.15 CPU) @ 27502.18/s (n=86511)
                   3.14 WALL ( 3.13 usr +  0.01 sys =  3.14 CPU) @ 27541.84/s (n=86511)
                   3.15 WALL ( 3.15 usr +  0.00 sys =  3.15 CPU) @ 27494.58/s (n=86511)
   oseq.to_array:  3.16 WALL ( 3.15 usr +  0.01 sys =  3.16 CPU) @ 40901.12/s (n=129308)
                   3.16 WALL ( 3.15 usr +  0.01 sys =  3.16 CPU) @ 40959.03/s (n=129308)
                   3.17 WALL ( 3.17 usr +  0.00 sys =  3.17 CPU) @ 40849.63/s (n=129308)
ccarray.to_array:  3.16 WALL ( 3.12 usr +  0.04 sys =  3.16 CPU) @ 24513.87/s (n=77389)
                   3.16 WALL ( 3.16 usr +  0.00 sys =  3.16 CPU) @ 24479.66/s (n=77389)
                   3.16 WALL ( 3.15 usr +  0.01 sys =  3.16 CPU) @ 24526.76/s (n=77389)
                    Rate     ccarray.to_array gen.to_array iter.to_array oseq.to_array
[...]

PR #472, opened 2025-01-29
add CCArray.of_iter

illusory0x0:
- #471

### DIFF
--- a/src/core/CCArray.mli
+++ b/src/core/CCArray.mli
@@ -240,6 +240,11 @@ val to_iter : 'a t -> 'a iter
     in modification of the iterator.
     @since 2.8 *)
 
+val of_iter : 'a iter -> 'a t
+(** [of_iter iter] builds a array from a given [iter].
+        In the result, elements appear in the same order as they did in the source [iter].
+     @since 3.15    *)
+
 val to_seq : 'a t -> 'a Seq.t
 (** [to_seq a] returns a [Seq.t] of the elements of an array [a].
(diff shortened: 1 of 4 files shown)


src/core/CCArray.mli
```diff
@@ -240,6 +240,11 @@ val to_iter : 'a t -> 'a iter
     in modification of the iterator.
     @since 2.8 *)
 
+val of_iter : 'a iter -> 'a t
+(** [of_iter iter] builds a array from a given [iter].
+        In the result, elements appear in the same order as they did in the source [iter].
+     @since 3.15    *)
```
c-cube, line 246: I normally use `@since NEXT_RELEASE` :)

c-cube (commented):
Looks useful, do you also want to add the updated benchmarks?

src/core/CCArray.mli
illusory0x0: thanks

illusory0x0:
> Looks useful, do you also want to add the updated benchmarks?

should we add [more benchmark](https://github.com/c-cube/ocaml-containers/issues/471#issuecomment-2621579760)?

when iter_len very small, use list as buffer is **faster** than other implementation.

c-cube:
lists might be faster for small sizes, but what about large sizes? does it get worse?

illusory0x0:
it would get worse.

look this [benchmark](https://github.com/c-cube/ocaml-containers/issues/471#issuecomment-2621579760) for more details

PR #480, opened 2025-10-27
fix: insert missing symbol into range doc comments

isti115:
The `^` character was missing from the exclusive range operator in two places.


c-cube:
Thank you!

PR #492, opened 2026-06-14
optimizations for KMP

c-cube:
alongside benchmarks.

### DIFF
--- a/benchs/run_benchs.ml
+++ b/benchs/run_benchs.ml
@@ -1636,6 +1636,159 @@ module Str = struct
     assert (CCList.equal CCInt.equal (mk_naive ()) (mk_current ()));
     B.throughputN 3 ~repeat [ "naive", mk_naive, (); "current", mk_current, () ]
 
+  (* Inlined reference implementations to have a fixpoint to compae
+      CCString with.
+
+     - [Classic_kmp] is a copy of CCString's old KMP: the
+       basic ("weak") failure table and the window-shift search loop.
+     - [New_kmp] is a bit optimized. *)
+  module Classic_kmp = struct
+    type t = {
+      failure: int array;
+      pat: string;
+    }
+
+    let compile pat =
+      let len = String.length pat in
+      match len with
+      | 0 -> { failure = [||]; pat }
+      | 1 -> { failure = [| -1 |]; pat }
+      | _ ->
+        let failure = Array.make len 0 in
+        failure.(0) <- -1;
+        let i = ref 2 and j = ref 0 in
+        while !i < len do
+          if CCChar.equal pat.[!i - 1] pat.[!j] then (
+            incr j;
+            failure.(!i) <- !j;
+            incr i
+          ) else if !j = 0 then (
+            failure.(!i) <- 0;
+            incr i
+          ) else
+            j := failure.(!j)
+        done;
+        { failure; pat }
+
+    let find { failure; pat } s =
+      let len = String.length s in
+      let i = ref 0 and j = ref 0 in
+      let pat_len = String.length pat in
+      while !j < pat_len && !i + !j < len do
+        if CCChar.equal s.[!i + !j] pat.[!j] then
+          incr j
+        else (
+          let fail_offset = failure.(!j) in
+          if fail_offset >= 0 then (
+            assert (fail_offset < !j);
+            i := !i + !j - fail_offset;
+            j := fail_offset
+          ) else (
+            j := 0;
+            incr i
+          )
+        )
+      done;
+      if !j = pat_len then
+        !i
+      else
+        -1
+  end
+
+  module New_kmp = struct
+    type t = {
+      failure: int array;
+      pat: string;
+    }
+
+    let compile pat =
+      let len = String.length pat in
+      match len with
+      | 0 -> { failure = [||]; pat }
+      | 1 -> { failure = [| -1 |]; pat }
+      | _ ->
+        let failure = Array.make len 0 in
+        failure.(0) <- -1;
+        let i = ref 2 and j = ref 0 in
+        while !i < len do
+          if CCChar.equal pat.[!i - 1] pat.[!j] then (
+            incr j;
+            failure.(!i) <- !j;
+            incr i
+          ) else if !j = 0 then (
+            failure.(!i) <- 0;
+            incr i
+          ) else
+            j := failure.(!j)
+        done;
+        (* strengthen: if the char after the border equals the one we'd be
+           retrying, skip straight to that border's (already-strengthened)
+           target instead of re-comparing it and failing again *)
+        for k = 1 to len - 1 do
+          let b = failure.(k) in
+          if b >= 0 && CCChar.equal pat.[k] pat.[b] then
+            failure.(k) <- failure.(b)
+        done;
+        { failure; pat }
+
+    let find { failure; pat } s =
+      let len = String.length s in
+      let i = ref 0 and j = ref 0 in
+      let pat_len = String.length pat in
+      while !j < pat_len && !i + !j < len do
+        if CCChar.equal s.[!i + !j] pat.[!j] then
+          incr j
+        else (
+          let fail_offset = failure.(!j) in
+          if fail_offset >= 0 then (
+            assert (fail_offset < !j);
+            i := !i + !j - fail_offset;
+            j := fail_offset
+          ) else (
+            (* no usable border: advance past the mismatch (NOT [incr i]) *)
+            i := !i + !j + 1;
+            j := 0
+          )
+        )
+      done;
+      if !j = pat_len then
+        !i
+      else
+        -1
+  end
+
+  (* a worst case test to show the difference between classic and new KMP,
+     with a needle [a^size] searched in a haystack made of [a^(size-1)b]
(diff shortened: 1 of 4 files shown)


tests/core/t_string.ml
```diff
@@ -47,8 +49,87 @@ eq' 6 (rfind ~sub:"a" "a1a234a");;
...
+(* Compare the KMP-based search against a naive reference, over a 2-char
+   alphabet so that needles actually occur in the haystack. This is a
+   two-sided check: it catches false negatives, wrong (non-leftmost/rightmost)
+   positions, and missed overlapping matches — none of which the properties
+   above (one-sided, large alphabet) can see. Needles can be longer than the
+   haystack (exercising the short-circuit) and repetitive (exercising the
+   failure table). *)
+let gen_ab ~maxlen =
+  Q.Gen.(
+    string_size (int_range 0 maxlen) ~gen:(fun st ->
```
FardaleM, line 60: Using `int_range 1 maxlen` would avoid needing the `Q.assume` in the test. Except if you want to test the search inside an empty haystack.

benchs/run_benchs.ml
```diff
@@ -1636,6 +1636,159 @@ module Str = struct
     assert (CCList.equal CCInt.equal (mk_naive ()) (mk_current ()));
     B.throughputN 3 ~repeat [ "naive", mk_naive, (); "current", mk_current, () ]
 
+  (* Inlined reference implementations to have a fixpoint to compae
```
FardaleM:
```suggestion
  (* Inlined reference implementations to have a fixpoint to compare
```

benchs/run_benchs.ml
```diff
@@ -1636,6 +1636,159 @@ module Str = struct
...
+          ) else
+            j := failure.(!j)
+        done;
+        { failure; pat }
+
+    let find { failure; pat } s =
+      let len = String.length s in
+      let i = ref 0 and j = ref 0 in
+      let pat_len = String.length pat in
+      while !j < pat_len && !i + !j < len do
```
FardaleM, line 1762: I wanted to suggest:
```suggestion
      while !j < pat_len && !i + pat_len <= len do
```
But running the benchmark show that it is 4% slower. No idea why. Same for `New_kmp`

benchs/run_benchs.ml
```diff
@@ -1636,6 +1636,159 @@ module Str = struct
...
+        else (
+          let fail_offset = failure.(!j) in
+          if fail_offset >= 0 then (
+            assert (fail_offset < !j);
+            i := !i + !j - fail_offset;
+            j := fail_offset
+          ) else (
+            j := 0;
+            incr i
+          )
```
FardaleM, line 1774:
```suggestion
          assert (fail_offset < !j);
          i := !i + !j - fail_offset;
          if fail_offset >= 0 then j := fail_offset;
```
If I understand the algorithm correctly, we have the following:
`fail_offset < 0` iff `fail_offset = -1` iff `j  = 0`.
Therefore, in the case `fail_offset <0`:
`i := !i + !j - fail_offset` correspond to `i := !i + 0 - (-1)` which is `incr i`
`j := 0` is not useful, as `j` is already `0`.

benchs/run_benchs.ml
```diff
@@ -1636,6 +1636,159 @@ module Str = struct
...
+            incr i
+          ) else
+            j := failure.(!j)
+        done;
+        (* strengthen: if the char after the border equals the one we'd be
+           retrying, skip straight to that border's (already-strengthened)
+           target instead of re-comparing it and failing again *)
+        for k = 1 to len - 1 do
+          let b = failure.(k) in
+          if b >= 0 && CCChar.equal pat.[k] pat.[b] then
```
FardaleM:
```suggestion
          if CCChar.equal pat.[k] pat.[b] then
```
`b >= 0` should always be true when `k>0`.

benchs/run_benchs.ml
```diff
@@ -1636,6 +1636,159 @@ module Str = struct
...
+          let fail_offset = failure.(!j) in
+          if fail_offset >= 0 then (
+            assert (fail_offset < !j);
+            i := !i + !j - fail_offset;
+            j := fail_offset
+          ) else (
+            (* no usable border: advance past the mismatch (NOT [incr i]) *)
+            i := !i + !j + 1;
+            j := 0
+          )
```
FardaleM, line 1833:
```suggestion
          assert (fail_offset < !j);
          i := !i + !j - fail_offset;
          if fail_offset >= 0 then j := fail_offset;

```
Same as before

benchs/run_benchs.ml
```diff
@@ -1636,6 +1636,159 @@ module Str = struct
...
+        )
+      done;
+      if !j = pat_len then
+        !i
+      else
+        -1
```
FardaleM, line 1839: I think with the comment above, `New_kmp.find` and `Classic_kmp.find` are identical. Maybe we can remove one and keep only one?

src/core/CCString.ml
```diff
@@ -45,11 +45,12 @@ type _ direction =
 (* we follow https://en.wikipedia.org/wiki/Knuth–Morris–Pratt_algorithm *)
 module Find = struct
   type 'a kmp_pattern = {
-    failure: int array;
+    mutable failure: int array; (* empty=not computed yet *)
```
FardaleM, line 48: I don't know how multicore works, so maybe I am saying something stupid.

By the introduction of a mutable field, is the use of pattern safe if shared between multiple domain? Should this be some kind of atomic field?

src/core/CCString.ml
```diff
@@ -92,8 +95,27 @@ module Find = struct
...
       (* Format.printf "{@[failure:%a, str:%s@]}@." CCFormat.(array int) failure str; *)
-      { failure; str }
+
+      (* strengthen the failure function: if falling back to a border would
+         re-compare a character equal to the one that just mismatched, skip
+         straight to that border's (already-strengthened) target instead.
+         Uses the direction-aware [get] so it works for both directions. *)
+      for k = 1 to len - 1 do
+        let b = failure.(k) in
+        if b >= 0 && CCChar.equal (get str k) (get str b) then
```
FardaleM:
```suggestion
        if CCChar.equal (get str k) (get str b) then
```
`b >= 0` should always be true when `k>0`.

src/core/CCString.ml
```diff
@@ -92,8 +95,27 @@ module Find = struct
...
+          failure.(k) <- failure.(b)
+      done;
+      failure
+
+  let kmp_compile_ : type a. dir:a direction -> string -> a kmp_pattern =
+   fun ~dir str -> { str; failure = [||]; dir }
+
+  let[@inline] kmp_get_ (type a) (f : a kmp_pattern) : int array =
+    if Array.length f.failure = 0 then
```
FardaleM, line 118: When the pattern `f.str` is empty, it makes that `failure` is always recomputed. As you forbid empty pattern now, it seems OK but in case it becomes possible in the future to have empty pattern, that might cause slowdown.

src/core/CCString.ml
```diff
@@ -107,23 +129,24 @@ module Find = struct
...
           i := !i + !j - fail_offset;
           j := fail_offset
         ) else (
-          (* beginning of pattern *)
-          j := 0;
-          incr i
+          (* no usable border: advance past the mismatch *)
+          i := !i + !j + 1;
+          j := 0
         )
```
FardaleM, line 127: Same as in the benchmark file. This can be simplied.

src/core/CCString.ml
```diff
@@ -107,23 +129,24 @@ module Find = struct
...
       )
```
FardaleM, line 154: The same simplification as in `run_benchs.ml` is possible here.

src/core/CCString.ml
```diff
@@ -150,16 +174,16 @@ module Find = struct
...
           i := !i + !j - fail_offset;
           j := fail_offset
         ) else (
-          (* beginning of pattern *)
-          j := 0;
-          incr i
+          (* no usable border: advance past the mismatch *)
+          i := !i + !j + 1;
+          j := 0
         )
```
FardaleM, line 163: The same here.

FardaleM (commented):
When giving an empty needle, `Find.compile` and `Find.rcompile` now raise `Invalid_argument`.
This behaviour is not yet documented in the `Split` module for `Split.{gen,list,seq,iter}(_copy)?` functions. Also, the documentation of CCStringLabel need the same changes as CCString.

Maybe this should be in the release note as breaking change. For most functions, it is more a bug fix, but for `String.find` and `String.mem` for example, the previous behaviour (`0` and `true`) were kind of making sense and could be legitimately used.

The main concern I have is with respect to the lazy construction of the `failure` table and if it works when sharing a pattern in multiple domain. But I don't know enough how multicore works to tell.

Otherwise, nice improvement! It was fun to review. Where does this idea come from?

tests/core/t_string.ml
c-cube: good catch. doing that for the needle only though.

src/core/CCString.ml
c-cube: good point, I documented the relevant parts and made it eager otherwise :)

